PR #2, opened 2025-09-22
VPD-932 Authorisation flow with backend integration

hmrc-spifix:


app/config/Module.scala
```diff
@@ -28,9 +29,11 @@ class Module extends AbstractModule {
     bind(classOf[DataRetrievalAction]).to(classOf[DataRetrievalActionImpl]).asEagerSingleton()
     bind(classOf[DataRequiredAction]).to(classOf[DataRequiredActionImpl]).asEagerSingleton()
 
-    // For session based storage instead of cred based, change to SessionIdentifierAction
-    bind(classOf[IdentifierAction]).to(classOf[AuthenticatedIdentifierAction]).asEagerSingleton()
+    // For session based storage instead of cred based, change to SessionIdentifyAction
```
ian-hmrc, line 32: We seem to have lost SessionIdentifierAction with the replacement of IdentifierAction.scala with IdentifyAction.scala. I think we can just remove this comment.

app/controllers/IndexController.scala
```diff
@@ -16,20 +16,26 @@
...
+  identify: IdentifyAction,
+  vapingDutyConnector: VapingDutyConnector,
+  view: IndexView
+)(implicit ec: ExecutionContext) extends FrontendBaseController with I18nSupport {
 
-  def onPageLoad(): Action[AnyContent] = identify { implicit request =>
-    Ok(view())
-  }
+  def onPageLoad(): Action[AnyContent] = identify.async(implicit request =>
+    vapingDutyConnector.ping().as(Ok(view()))
```
ian-hmrc, line 39: Bringing in cats just to do this seems a bit heavyweight. Shall we wait until we have real need of it before bringing it in?

hmrc-spifix: cats is great for formalising common functional patterns in a ubiquitous language, this is one example

app/controllers/actions/IdentifyAction.scala
```diff
@@ -0,0 +1,98 @@
...
+    case _                  => Redirect(routes.UnauthorisedController.onPageLoad())
+  }
+
+  private def getApprovalId(enrolments: Enrolments): Either[String, String] =
+    enrolments
+      .enrolments
+      .find(_.key == config.enrolmentServiceName)
+      .flatMap(_.getIdentifier(config.enrolmentIdentifierKey))
+      .map(_.value)
+      .toRight("Unable to retrieve VPPAID from enrolments")
```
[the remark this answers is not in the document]
hmrc-spifix: I see it now, addressed

app/controllers/IndexController.scala
ian-hmrc: I've just been reading up a little more on Cats. I've not used it in anger before and while I understand having these abstractions should make the code simpler, it is dependent on those working on the code understanding those abstractions well. I want to be sure that the whole tech team will be happy to own and work on the service, so I'd like to see who we recruit and what their familiarity with it is. We also have to consider this code will be handed to live services eventually, and keeping to vanilla Scala means they should have an easier time with it. For these reasons, and the fact that this case has a trivial implementation in vanilla Scala, I think we should not introduce it at this point.

This doesn't mean that we should never introduce cats. If we have a situation in the future where cats brings clear benefits, we can consider it at that time, and with with knowing who is on the team and their level of familiarity with it.

ian-hmrc (approved):
LGTM

PR #72, opened 2026-03-19
VPD-1470: lint: use ScalaFix in favour of Scalafmt

AsadHumayun:
- showcase potential impact from using scalafix
- use scalafix
- update sbt aliases
- use `sbt scalafixAll` to lint, alias `sbt lint` can also be used

43867aab5478d87db73c9f231117e376d5236111 applies `sbt lint`. These changes are mostly import organisations. They _should_ not be changing anything.

todo: apply these changes to backend, finance, account, stubs


AsadHumayun:
Changing config to also force explicit type annotations

PR #74, opened 2026-03-26
VPD-1472: Update instances of VpdId, InternalId and CredentialId to b…

ghousedev:
…e domain types rather than Strings

### DIFF
--- a/app/config/FrontendAppConfig.scala
+++ b/app/config/FrontendAppConfig.scala
@@ -17,7 +17,7 @@
 package config
 
 import com.google.inject.{Inject, Singleton}
-import models.{CredentialId, InternalId, VpdId}
+import models.identifiers.{CredentialId, InternalId, VpdId}
 import play.api.Configuration
 import play.api.i18n.Lang
 import play.api.mvc.RequestHeader
--- a/app/config/FrontendAppConfig.scala
+++ b/app/config/FrontendAppConfig.scala
@@ -82,7 +82,7 @@ class FrontendAppConfig @Inject() (configuration: Configuration, servicesConfig:
   def cpUserAnswersGetUrl(vpdId: VpdId): String               = s"$contactPreferenceBaseUrl/user-answers/$vpdId"
   def cpUserAnswersUrl: String                                = s"$contactPreferenceBaseUrl/user-answers"
   def cpUserAnswersKeepAliveUrl: String                       = s"$contactPreferenceBaseUrl/keep-alive"
-  def cpUserAnswersClearUrl(userId: InternalId): String       = s"$contactPreferenceBaseUrl/user-answers/clear/$userId"
+  def cpUserAnswersClearUrl(internalId: InternalId): String   = s"$contactPreferenceBaseUrl/user-answers/clear/$internalId"
   def cpGetEmailVerificationUrl(credId: CredentialId): String = s"$contactPreferenceBaseUrl/get-email-verification/$credId"
   def cpSubmitContactPreferencesUrl(vpdId: VpdId): String     = s"$contactPreferenceBaseUrl/submit-preferences/$vpdId"
   def getSubscriptionUrl(vpdId: VpdId): String                = s"$contactPreferenceBaseUrl/get-preferences/$vpdId"
--- a/app/config/FrontendAppConfig.scala
+++ b/app/config/FrontendAppConfig.scala
@@ -17,6 +17,7 @@
 package config
 
 import com.google.inject.{Inject, Singleton}
+import models.{CredentialId, InternalId, VpdId}
 import play.api.Configuration
 import play.api.i18n.Lang
 import play.api.mvc.RequestHeader
@@ -38,8 +39,8 @@ class FrontendAppConfig @Inject() (configuration: Configuration, servicesConfig:
   def feedbackUrl(implicit request: RequestHeader): String =
     s"$contactHost/contact/beta-feedback?service=$contactFormServiceIdentifier&backUrl=${host + request.uri}"
 
-  val enrolmentServiceName = configuration.get[String]("enrolment.serviceName")
-  val enrolmentIdentifierKey = configuration.get[String]("enrolment.identifierKey")
+  val enrolmentServiceName: String          = configuration.get[String]("enrolment.serviceName")
+  val enrolmentIdentifierKey: String        = configuration.get[String]("enrolment.identifierKey")
 
   val loginUrl: String                       = configuration.get[String]("urls.login")
   val loginContinueUrl: String               = configuration.get[String]("urls.loginContinue")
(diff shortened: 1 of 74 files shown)


it/test/repositories/EnrolmentSessionRepositoryISpec.scala
```diff
@@ -83,7 +84,7 @@ class EnrolmentSessionRepositoryISpec
 
         insert(userAnswers).futureValue
 
-        val result         = repository.get(userAnswers.id).futureValue
+        val result         = repository.get(InternalId(userAnswers.id)).futureValue
```
ian-hmrc: I can see a lot of this wrapping going on. Would it be possible to provide a method/property like `userAnswers.internalId` to do this? Or would that get caught up in the serialisation?

app/config/FrontendAppConfig.scala
```diff
@@ -76,17 +77,15 @@ class FrontendAppConfig @Inject() (configuration: Configuration, servicesConfig:
...
-  def getSubscriptionUrl(vpdId: String): String             = s"$contactPreferenceBaseUrl/get-preferences/$vpdId"
+  
+  private val contactPreferenceBaseUrl                        = s"$contactPreferencesHost/vaping-duty-account"
+  def cpUserAnswersGetUrl(vpdId: VpdId): String               = s"$contactPreferenceBaseUrl/user-answers/$vpdId"
+  def cpUserAnswersUrl: String                                = s"$contactPreferenceBaseUrl/user-answers"
+  def cpUserAnswersKeepAliveUrl: String                       = s"$contactPreferenceBaseUrl/keep-alive"
+  def cpUserAnswersClearUrl(userId: InternalId): String       = s"$contactPreferenceBaseUrl/user-answers/clear/$userId"
+  def cpGetEmailVerificationUrl(credId: CredentialId): String = s"$contactPreferenceBaseUrl/get-email-verification/$credId"
+  def cpSubmitContactPreferencesUrl(vpdId: VpdId): String     = s"$contactPreferenceBaseUrl/submit-preferences/$vpdId"
+  def getSubscriptionUrl(vpdId: VpdId): String                = s"$contactPreferenceBaseUrl/get-preferences/$vpdId"
```
ian-hmrc, line 88: I think this really helps ensure we're using the right Id for all these URLs. 👍

app/repositories/EnrolmentUserAnswersRepository.scala
```diff
@@ -53,9 +54,9 @@ class EnrolmentUserAnswersRepository @Inject()(
 
   implicit val instantFormat: Format[Instant] = MongoJavatimeFormats.instantFormat
 
-  private def byId(id: String): Bson = Filters.equal("_id", id)
+  private def byId(id: InternalId): Bson = Filters.equal("_id", id.toString)
```
ian-hmrc, line 57: Might this function now be better named `byInternalId()`?

app/repositories/EnrolmentUserAnswersRepository.scala
```diff
@@ -53,9 +54,9 @@ class EnrolmentUserAnswersRepository @Inject()(
...
 
-  def keepAlive(id: String): Future[Boolean] =
+  def keepAlive(id: InternalId): Future[Boolean] =
```
ian-hmrc, line 59: Could these parameters be renamed from `id` to `internalId`?

app/connectors/SubmitPreferencesConnector.scala
```diff
@@ -36,8 +37,8 @@ class SubmitPreferencesConnector @Inject() (
     extends HttpReadsInstances
     with Logging {
 
-  def submitContactPreferences(contactPreferenceSubmission: PaperlessPreferenceSubmission, vpdId: String)(implicit
-    hc: HeaderCarrier
+  def submitContactPreferences(contactPreferenceSubmission: PaperlessPreferenceSubmission, vpdId: VpdId)(implicit
+                                                                                                         hc: HeaderCarrier
```
ian-hmrc: The indentation has gone a little awry here. Could this be a bit more readable? Split params to separate lines maybe?

ghousedev: Tidied that up

it/test/repositories/EnrolmentSessionRepositoryISpec.scala
ian-hmrc: OK, I can see you've pulled out a property for that. I wonder if the deeper thing going on here is that we have a relationship between the internal id and the user answers (key and value), but the user answers also contain the key (the internal Id). Again, this might be sorted if we separated our representations for mongo from those in our domain, but I wouldn't do that on top of this PR. Let's ponder this idea and see where it goes.

app/connectors/UserAnswersConnector.scala
```diff
@@ -52,7 +52,7 @@ class UserAnswersConnector @Inject() (config: FrontendAppConfig, implicit val ht
       .execute[Either[UpstreamErrorResponse, UserAnswers]]
   }
 
-  def keepAlive(vpdId: String)(implicit hc: HeaderCarrier): Future[Either[UpstreamErrorResponse, Unit]] =
+  def keepAlive(userId: InternalId)(implicit hc: HeaderCarrier): Future[Either[UpstreamErrorResponse, Unit]] =
```
ian-hmrc: The use of `vpdId` in the old version is a little worrying here, but I think it was just a copy 'n' paste error. I've checked and the correction to `userId` seems right. Might just be worth double-checking though.

BTW, was there a reason use `userId` over `internalId` for the parameter names here? I think adding `InternalId` has made things clearer, and maybe the parameter name could follow that? I think the clarity would make it clear that it's no other form of user id.

app/controllers/actions/ApprovedVapingManufacturerAuthAction.scala
```diff
@@ -79,7 +80,8 @@ class ApprovedVapingManufacturerAuthActionImpl @Inject()(override val authConnec
         }
 
         identifiers match {
-          case Right((internalId, groupId, approvalId, credId)) => block(IdentifierRequest(request, approvalId, groupId, internalId, credId))
+          case Right((internalId, groupId, approvalId, credId)) =>
+            block(IdentifierRequest(request, VpdId(approvalId), groupId, InternalId(internalId), CredentialId(credId)))
```
ian-hmrc: When we chatted, you suggested that groupId could also be wrapped. I think that's worth doing.

app/controllers/actions/ClaimEnrolmentAuthAction.scala
```diff
@@ -78,7 +79,7 @@ class EnrolmentClaimAuthActionImpl @Inject()(override val authConnector: AuthCon
         }
 
         identifiers match {
-          case Right((internalId, groupId, optApprovalId)) => block(NoEnrolmentIdentifierRequest(request, optApprovalId, groupId, internalId))
+          case Right((internalId, groupId, optApprovalId)) => block(NoEnrolmentIdentifierRequest(request, optApprovalId, groupId, InternalId(internalId)))
```
ian-hmrc: `groupId` could also be wrapped here.

AsadHumayun (approved):

PR #75, opened 2026-03-27
VPD-1475: Refactor HowToBeContactedController

ghousedev:


app/controllers/contactPreference/HowToBeContactedController.scala
craigemmerson, line 69: are we handling any potential Try failures somewhere ? what if this fails ? 

perhaps a .recover on the for yield and redirect to recovery controller if we fail ? worth a thought.

craigemmerson: just an idea, we could make this a one liner with 

private def prepareForm(ua: UserAnswers) =
  ua.get(HowToBeContactedPage).fold(form)(form.fill)

as we are just dealing with an option here. depends on readable vs tidy preference

ian-hmrc (approved):
LGTM :)

app/controllers/contactPreference/HowToBeContactedController.scala
ghousedev: Implemented the suggestion

PR #79, opened 2026-04-09
VPD-1508: Fix email wrap bug

AsadHumayun:
- fixes the bug to wrap on all email fields that I could find in our service
- ~~I cannot test it at the end of the journey (the confirmation part)~~ It works fine


AsadHumayun:
Have QA approval to merge this once it gets approved

ghousedev (approved):
👍

PR #80, opened 2026-04-10
VPD-1511: Refactor FE

ghousedev:


craigemmerson (dismissed):

ian-hmrc:
Could `CheckSignedInAction` be moved under the `enrolment` package? It is only used for the Enrolment pages.

ian-hmrc:
Would the `models.audit.contactPreference` make more sense as `models.contactPreference.audit`?

Also, while this is visible, [Scala package names should be all lower case](https://docs.scala-lang.org/tour/packages-and-imports.html#creating-a-package). I'm not sure what it should be in this case, as `contactpreference` is not very legible.

craigemmerson (approved):

PR #87, opened 2026-05-07
VPD-1528: Build 'view individual returns'

AsadHumayun:


app/connectors/returns/GetReturnsConnector.scala
```diff
@@ -0,0 +1,69 @@
...
+    response match {
+      case Right(response) =>
+        Try {
+          response.json.as[ReturnDisplayResponse]
+        } match {
+          case Success(returnsData) => Future.successful(returnsData)
+          case Failure(_) =>
+            logger.warn(parsingError)
+            Future.failed(InternalServerException(parsingError))
+        }
```
ian-hmrc: Just a thought here, but this function is doing 2 jobs and they could be split into individual functions. It is converting the `Either` into a `Future` and parsing the JSON response. I can show you the refactoring if you like.

app/controllers/actions/returns/ReturnsDataRequiredAction.scala
```diff
@@ -32,7 +32,7 @@ class ReturnsDataRequiredActionImpl @Inject()(implicit val executionContext: Exe
       case None =>
         Future.successful(Left(Redirect(routes.JourneyRecoveryController.onPageLoad())))
       case Some(data) =>
-        Future.successful(Right(ReturnsDataRequest(request.request, request.enrolmentVpdId, request.internalId, request.credId, data)))
+        Future.successful(Right(ReturnsDataRequest(request.request, request.enrolmentVpdId, request.internalId, request.credId, Option(data.periodKey).flatten.getOrElse(""), data)))
```
ian-hmrc: This is quite a long line. Could the params be put on separate lines?

Also, isn't `Option(data.periodKey).flatten.getOrElse("")` more complex than it needs to be? `data.periodKey` is an `Option` and so should never be null. I've run the tests with this change, but the tests fail because the test data has `data.periodKey` as null. I've not been able to dig in enough to find out why. I think if we fix the test data then this could be simplified.

app/controllers/returns/submit/BeforeYouStartController.scala
```diff
@@ -17,26 +17,43 @@
...
+      val ua = request.userAnswers match {
+        case Some(existingUa) if Option(existingUa.periodKey).flatten.contains(pk) =>
+          existingUa
+        case _ =>
+          ReturnsUserAnswers.getEmptyReturnsUA(request.enrolmentVpdId, pk)
+      }
+
+      sessionRepository.set(ua.copy(periodKey = Some(pk))).map(_ =>
+        Ok(view(BeforeYouStartViewModel())).withSession(session)
+      )
```
ian-hmrc, line 56: Hmm... This is the only place where the period key is stored on the session, so to switch between return submissions the user would have to navigate via the Before You Start page. What happens if the user has 2 tabs open on one browser with a different return in each? Are they able to interleave filling-out each return, or is the session share between both?

app/controllers/returns/view/ViewIndividualReturnController.scala
```diff
@@ -0,0 +1,52 @@
...
+                                       returnsRepository: ReturnsUserAnswersConnector,
+                                       getData: ReturnsDataRetrievalAction
+                                     )(using ExecutionContext) extends FrontendBaseController with I18nSupport {
+
+  def onPageLoad(periodKey: String): Action[AnyContent] = (identify andThen returnsEnabled andThen getData).async {
+    implicit request =>
+      connector.getReturn(periodKey, vpdId = request.enrolmentVpdId)
+        .map { returnData =>
+          val ua = request.userAnswers.fold(ReturnsUserAnswers.getEmptyReturnsUA(request.enrolmentVpdId, returnData.success.chargeDetails.get.periodKey))(ua => ua)
+          returnsRepository.set(ua.copy(periodKey = Some(periodKey)))
```
ian-hmrc: Why do the Submit Return User Answers need resetting here when viewing an individual return? Could this be removed?

ian-hmrc: If it is required, could something be done about the long line?

app/services/returns/SubmitReturnService.scala
```diff
@@ -38,20 +39,45 @@ class SubmitReturnService @Inject()(submitReturnConnector: SubmitReturnConnector
...
-      VapingProductsProduced(Seq.empty, Seq.empty),
-      TotalDutyDue(totalInMl, zeroValue, zeroValue, zeroValue, zeroValue, totalDue)
+    val vapingProductsProduced = dutyDeclared match {
+      case false  =>
+        VapingProductsProduced(nilReturn = Seq(NilReturn(vapingProductsProduced = "0")), regularReturn = Seq())
+      case _      =>
+        VapingProductsProduced(nilReturn = Seq(), regularReturn = Seq(RegularReturn(
+          taxType = "301", dutyRate = dutyRate, amountProducedLiquid = totalDue, dutyDue = dutyDue
+        )))
+    }
```
ian-hmrc, line 62: Could this match be simplified down to an if expression (my IDE is suggesting to do that)? Also, would introducing default parameters of empty sequences for the properties of `VapingProductsProduced` make this simpler?

app/services/returns/SubmitReturnService.scala
```diff
@@ -38,20 +39,45 @@ class SubmitReturnService @Inject()(submitReturnConnector: SubmitReturnConnector
...
     val totalInMl = ua.get(EnterDutyAmountPage).fold(BigDecimal(0))(value => BigDecimal(value))
 
     // Temp value
     val zeroValue = BigDecimal(0)
 
     // Will need to either get or pass the period key here
-    val periodKey = "26AF"
+    val periodKey = Option(ua.periodKey).flatten.fold("")(x => x)
 
     // Will need to enhance this much more
```
ian-hmrc: I agree. It is also starting to look quite procedural. It will likely be worth waiting until more of this is present but I think some functions might be extracted from this once the shape is clearer.

app/services/returns/SubmitReturnService.scala
```diff
@@ -38,20 +39,45 @@ class SubmitReturnService @Inject()(submitReturnConnector: SubmitReturnConnector
...
+      case false  =>
+        VapingProductsProduced(nilReturn = Seq(NilReturn(vapingProductsProduced = "0")), regularReturn = Seq())
+      case _      =>
+        VapingProductsProduced(nilReturn = Seq(), regularReturn = Seq(RegularReturn(
+          taxType = "301", dutyRate = dutyRate, amountProducedLiquid = totalDue, dutyDue = dutyDue
+        )))
+    }
+
+    val totalDutyDueVapingProducts  = vapingProductsProduced.regularReturn.headOption
+      .fold(zeroValue)(regularReturn => regularReturn.dutyDue)
```
ian-hmrc: Could this be simplified to just `if (dutyDeclared) dutyDue else zeroValue`? I think it would be more readable as it is closer to the domain.

app/viewmodels/returns/view/ViewIndividualReturnViewModel.scala
```diff
@@ -0,0 +1,170 @@
...
+
+    val totalDuty = success.totalDutyDue
+      .fold(currencyFormat(BigDecimal(0)))(td => currencyFormat(td.totalDutyDue))
+
+    val periodKey = returnsData.success.chargeDetails.get.periodKey
+
+    val year = s"20${periodKey.take(2)}"
+
+    val monthFromPeriodKey = PeriodKeys.fromEtmpMonthString(periodKey.takeRight(2))
+      .getOrElse(LocalDate.ofInstant(Instant.now(), ZoneId.systemDefault()).getMonth)
```
ian-hmrc: The return has `chargeDetails.periodFrom` and `chargeDetails.periodTo`. We could use these to derive the year and month more easily than trying to interpret it from the period key. That would allow us to remove the functions in `PeriodKeys`.

app/services/returns/SubmitReturnService.scala
```diff
@@ -38,20 +39,45 @@ class SubmitReturnService @Inject()(submitReturnConnector: SubmitReturnConnector
...
-    ReturnCreateRequest(
-      periodKey,
-      VapingProductsProduced(Seq.empty, Seq.empty),
-      TotalDutyDue(totalInMl, zeroValue, zeroValue, zeroValue, zeroValue, totalDue)
+    val vapingProductsProduced = dutyDeclared match {
+      case false  =>
+        VapingProductsProduced(nilReturn = Seq(NilReturn(vapingProductsProduced = "0")), regularReturn = Seq())
+      case _      =>
+        VapingProductsProduced(nilReturn = Seq(), regularReturn = Seq(RegularReturn(
+          taxType = "301", dutyRate = dutyRate, amountProducedLiquid = totalDue, dutyDue = dutyDue
```
ian-hmrc: The tax type for VPD will be 641 (see data items v2.7).  Also, though the frontend will interact with the user in millilitres, the volumes submitted to the API will be in litres, we should be careful with the units we are using here. Maybe we can use some types to ensure we're using ml and L correctly?

app/controllers/actions/returns/ReturnsDataRequiredAction.scala
ghousedev: Yes it is quite long I will make that change.


As discussed yesterday we plan to introduce a domain type for the period key, this will alleviate this issue and a couple of other comments further down around the period key and it's uses.

app/services/returns/SubmitReturnService.scala
ghousedev: Have moved to a config value and updated.

app/controllers/returns/view/ViewIndividualReturnController.scala
ghousedev: It is not required, would have been a left over from the iterations we tried to use the period key.

app/viewmodels/returns/view/ViewIndividualReturnViewModel.scala
ghousedev: Have changed to get the month/year from periodFrom.

app/controllers/actions/returns/ReturnsDataRequiredAction.scala
ian-hmrc: The type is not the problem. It's the test data. Tests like `EnterDutyAmountControllerSpec` fail if this change is made because the test data used is not setup correctly. The `returnUserAnswers` in `TestData` has a null `periodKey` because the `optPeriodKey` used is initialised further down the file and has not been initialised yet. Moving `optPeriodKey` above `returnUserAnswers` allows this to work correctly.

app/connectors/returns/GetReturnsConnector.scala
ian-hmrc: I can see you've made a change here, but it's only half of what I was thinking. My suggestion is to separate them into independent transformations:
```
      ...
      .flatMap(handleErrors)
      .flatMap(parseJson)

  private def handleErrors(response: Either[UpstreamErrorResponse, HttpResponse]): Future[HttpResponse] =
    response match {
      case Right(response) => Future.successful(response)
      case Left(error) =>
        logger.warn(s"Unexpected response from VPD return get API. Status: ${error.statusCode}")
        Future.failed(InternalServerException("Failed to get VPD return"))
    }

  private def parseJson(response: HttpResponse) =
    Try {
      response.json.as[ReturnDisplayResponse]
    } match {
      case Success(returnsData) => Future.successful(returnsData)
      case Failure(_) =>
        logger.warn(parsingError)
        Future.failed(InternalServerException(parsingError))
    }
```

app/services/returns/SubmitReturnService.scala
ian-hmrc: I still think the default params make the readability here, while we're stuck with this representation.

app/viewmodels/returns/view/ViewIndividualReturnViewModel.scala
ian-hmrc: The `fromDisplayName()` function in `PeriodKeys` is also unused and untested. Could this be removed too?

app/services/returns/SubmitReturnService.scala
```diff
@@ -38,20 +41,44 @@ class SubmitReturnService @Inject()(submitReturnConnector: SubmitReturnConnector
...
+
 
     // Will need to either get or pass the period key here
-    val periodKey = "26AF"
+    val periodKey = Option(ua.periodKey).flatten.fold("")(x => x)
 
     // Will need to enhance this much more
     val totalDue = totalInMl - zeroValue
+    val dutyRate = BigDecimal("2.2")
+    val dutyDue = (totalDue * dutyRate).setScale(2, BigDecimal.RoundingMode.HALF_UP)
```
ian-hmrc: This rounding is not correct. From the Return Data Items v2.7 for VPD-DRM-R007 'Rounded down in to the nearest penny'. Any rounding errors should go in the manufacturers' favour.

app/services/returns/SubmitReturnService.scala
ian-hmrc: Can we do something to convert the ml to L, even if it's just dividing by 1000 and naming variables with units? I think this is a really easy one to forget and it would be good to leave at least some clue for when we're back in here to re-do the calculation.

app/controllers/returns/submit/BeforeYouStartController.scala
ghousedev: Multiple tabs can cause issues, we do need to look at this.  Have a ticket in next sprint we can investigate in.

app/controllers/returns/submit/BeforeYouStartController.scala
```diff
@@ -17,26 +17,43 @@
...
+      val session = request.session + ("periodKey" -> pk)
+
+      val ua = request.userAnswers match {
+        case Some(existingUa) if existingUa.periodKey.contains(pk) =>
+          existingUa
+        case _ =>
+          ReturnsUserAnswers.getEmptyReturnsUA(request.enrolmentVpdId, pk)
+      }
+
+      sessionRepository.set(ua.copy(periodKey = pk)).map(_ =>
```
ian-hmrc: Just a small one, but I believe the `periodKey` is always set on the `ua` here. I think we don't need to copy them to set it, right?

app/controllers/returns/submit/BeforeYouStartController.scala
```diff
@@ -17,26 +17,43 @@
...
+                                     )(using ExecutionContext) extends FrontendBaseController with I18nSupport {
 
-  def onPageLoad: Action[AnyContent] = (identify andThen returnsEnabledAction) {
+  def onPageLoad(): Action[AnyContent] = (identify andThen returnsEnabledAction andThen getData).async {
     implicit request =>
+      val pk = request.getQueryString("period").getOrElse("")
+      val session = request.session + ("periodKey" -> pk)
+
+      val ua = request.userAnswers match {
+        case Some(existingUa) if existingUa.periodKey.contains(pk) =>
```
ian-hmrc: Is there a reason that we're using `contains` here? Could we not just test for equality?

test/controllers/returns/submit/BeforeYouStartControllerSpec.scala
```diff
@@ -17,19 +17,35 @@
...
 
   "BeforeYouStart Controller" - {
 
     "must return OK and the correct view for a GET" in {
-      val application = applicationBuilder(returnsUserAnswers = Some(returnsUserAnswers)).build()
+
+      val mockService = mock[ReturnsUserAnswersService]
+
+      val application = applicationBuilder(returnsUserAnswers = Some(returnsUserAnswers))
+        .overrides(bind[ReturnsUserAnswersService].to(mockService))
```
ian-hmrc, line 42: This test is where the `UserAnswers` already has a period key, so this is for the scenario where the user is returning to the submission after a break.

Could we add test for the scenario where the user is starting the return for the first time? I know there's probably not much that we can assert differently but capturing that intent would be good.

ian-hmrc: Actually, with a little further thought, would it be possible to only initialise the user answers if they are not already correctly setup? This would be less wasteful and allow the distinction to be made by these 2 tests.

app/controllers/returns/submit/BeforeYouStartController.scala
ghousedev: Made the change

app/controllers/returns/submit/BeforeYouStartController.scala
ghousedev: Yes I can remove the copy.

test/controllers/returns/submit/BeforeYouStartControllerSpec.scala
ian-hmrc: Cool. I can see the extra test. We can pick up with the controller changes when we make the enhancements around the period key.

ian-hmrc (approved):
LGTM :)

PR #92, opened 2026-05-22
VPD-1661: Update confirmation page

ghousedev:


app/views/returns/submit/ConfirmationEmailView.scala.html
```diff
@@ -45,10 +45,14 @@
     @p(Seq(
         messages("returns.confirmation.p.weHave"),
         HtmlContent(s" <span class=$boldFontWeight>${vm.email}</span>"),
-        messages("returns.confirmation.p.confirming", vm.date),
+        messages("returns.confirmation.p.confirming", vm.date)
     ))
 
-    @p(Seq(HtmlContent(link(id = "view-return", href = "", text = messages("returns.confirmation.link.view")))))
+    @p(Seq(HtmlContent(link(
```
craigemmerson, line 51: to keep in line with the rest of the service could we look at moving link logic to the VM please ?

craigemmerson (approved):

PR #93, opened 2026-05-26
VPD-1658: Duty rate by month

ghousedev:


app/models/returns/DutyRate.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import java.time.LocalDate
+
+final case class DutyRate(
+  startDate: LocalDate,
+  endDate: LocalDate,
+  ratePencePerMl: Int
+) {
+  def isValidFor(date: LocalDate): Boolean =
+    !date.isBefore(startDate) && !date.isAfter(endDate)
```
ian-hmrc: Would it be better to use a `Range.inclusive()` so it could be checked with `dutyPeriod.contains(date)`?

ian-hmrc: Actually, Range is integer only, but maybe there's another type out there, or maybe we could create our own `DateRange`?

test/config/DutyRateConfigSpec.scala
```diff
@@ -0,0 +1,178 @@
...
+
+import base.SpecBase
+import com.typesafe.config.ConfigFactory
+import play.api.Configuration
+
+import java.time.LocalDate
+
+class DutyRateConfigSpec extends SpecBase {
+
+  private val validConfig = Configuration(ConfigFactory.parseString(
```
ian-hmrc: Could this be inlined as it is only used once and this would be more in-keeping with the other tests that setup their own data.

app/config/DutyRateConfig.scala
```diff
@@ -0,0 +1,74 @@
...
+    
+    // Check for gaps and overlaps
+    rates.sliding(2).foreach {
+      case Seq(current, next) =>
+        val dayAfterCurrentEnd = current.endDate.plusDays(1)
+        require(
+          dayAfterCurrentEnd.isEqual(next.startDate),
+          s"Gap or overlap detected between periods: ${current.endDate} and ${next.startDate}"
+        )
+      case _ => RuntimeException("Duty rates misconfigured")
```
ian-hmrc: This exception is not being thrown, just created and then dropped as the function goes on to do other procedural validation and then return `Unit`. This is for the case when `rates.sliding(2)` has fewer rates than the window (2), so when there is only 1 rate (which will be the initial case). We get away with this (I added the missing test for this case and it does pass). I think I would prefer it if we didn't rely on `require()` and exceptions for the validation, and instead had more modular functions for this validation that could be unit tested individually (one for 'validate each rate', one for 'check gaps and overlaps', etc.), likely returning an `Either`. Using TDD to drive the design of this would be useful.

app/config/DutyRateConfig.scala
```diff
@@ -0,0 +1,74 @@
...
+        endDate = LocalDate.parse(rateConfig.get[String]("end-date")),
+        ratePencePerMl = rateConfig.get[Int]("rate-pence-per-ml")
+      )
+    }.sortBy(_.startDate)
+    
+    validateRates(parsedRates)
+    parsedRates
+  }
+  
+  private def validateRates(rates: Seq[DutyRate]): Unit = {
```
ian-hmrc: This validation will only take place when a manufacturer tries to submit a return and calculates some duty using the rate, at which point they would see a service failure. Ideally this validation would take place when the application is deployed (maybe even preventing the app from starting up?), as this would be when the person at HMRC had updated the rates incorrectly, and the issue could be corrected before affecting any customers. Can this be done?

test/config/DutyRateConfigSpec.scala
```diff
@@ -0,0 +1,178 @@
...
+          |]
+          |""".stripMargin))
+
+      val config = new DutyRateConfig(unsortedConfig)
+      
+      config.rates.head.startDate mustBe LocalDate.of(2026, 1, 1)
+      config.rates(1).startDate mustBe LocalDate.of(2027, 1, 1)
+    }
+
+    "must fail when no rates are configured" in {
```
ian-hmrc, line 62: These tests, especially for validation failures, are quite large and integrated. I would expect each validation to be independent and individually testable. Can we make these more modular and test more locally?

test/config/DutyRateConfigSpec.scala
```diff
@@ -0,0 +1,178 @@
...
+      val config = new DutyRateConfig(validConfig)
+      
+      config.rates must have size 2
+      config.rates.head.startDate mustBe LocalDate.of(2026, 1, 1)
+      config.rates.head.endDate mustBe LocalDate.of(2026, 12, 31)
+      config.rates.head.ratePencePerMl mustBe 22
+      config.rates(1).ratePencePerMl mustBe 30
+    }
+
+    "must sort rates by start date" in {
```
ian-hmrc: Could a test be added for a single duty rate? This is how the service will start out and we should be sure that that case works.

app/config/DutyRateConfig.scala
craigemmerson: Have you ran the app to verify this ? we ran this app with sam yesterday with invalid date ranges and the app did not start which proved this. we took a screen shot of it: edit - I thought we took a screen shot I don't know if sam did. but we tested this part in front of him to fail app start up

conf/application.conf
```diff
@@ -131,3 +131,20 @@ features {
 taxType {
   vpd = "641"
 }
+duty-rates = [
```
ian-hmrc, line 141: Could we add some comments here to help anybody who is adding another rate to do the right thing? It will be stuff to try to make sure they don't get tripped-up by the validation:

- the date ranges need to be contiguous
- rates must be positive
- current date must be covered
- rates can be added before they come into effect
- final date should stretch far into the future (e.g. 9999-12-31) so that rate effectively applies forever.

We should also inform them that making mistakes will prevent the app from launching.

Likely the process for adding another rate will be:

- duplicate the final entry
- end the previous rate a day before the new rate comes into force
- start the new entry on its start date
- set the new rate.

I presume also that there will be a PR for **app-config-common** with the initial 22p/ml from October (and maybe app-config-staging with test values?). The comments should be duplicated there too.

app/services/returns/ObligationService.scala
```diff
@@ -0,0 +1,52 @@
...
+        .map(_.obligationDetails)
+        .find(_.periodKey == periodKey)
+    }
+  
+  def getDutyRateForPeriod(vpdId: VpdId, periodKey: String)
+                          (using HeaderCarrier): Future[Option[BigDecimal]] =
+    getObligationByPeriodKey(vpdId, periodKey).map { obligationOpt =>
+      obligationOpt.map { obligation =>
+        val rate = dutyRateService.getRateForDate(obligation.iCFromDate)
+        BigDecimal(rate) / 100
```
ian-hmrc: I think we need types to keep clear the units we are dealing with. The `rate` starts out as p/ml, but the return value is £/ml, right?

app/controllers/returns/submit/CheckYourAnswersController.scala
```diff
@@ -35,18 +36,27 @@ class CheckYourAnswersController @Inject()(
...
 
-  def onPageLoad: Action[AnyContent] = (identify andThen returnsEnabled andThen getData andThen requireData) { implicit request =>
-    val vm = CheckYourAnswersViewModel(request.userAnswers)
-    Ok(view(vm))
+  def onPageLoad: Action[AnyContent] = (identify andThen returnsEnabled andThen getData andThen requireData).async { implicit request =>
+    obligationService.getDutyRateForPeriod(
+      VpdId(request.userAnswers.vpdId),
+      request.userAnswers.periodKey
+    ).map { dutyRateOpt =>
+      val dutyRate = dutyRateOpt.getOrElse(BigDecimal(0))
```
ian-hmrc, line 49: I don't think we should substitute zero here when we don't have a duty rate. This should be a failure from the controller. Would a for comprehension be better here?

app/controllers/returns/view/ViewIndividualReturnController.scala
```diff
@@ -32,16 +33,20 @@ class ViewIndividualReturnController @Inject()(
...
 
   def onPageLoad(periodKey: String): Action[AnyContent] = (identify andThen returnsEnabled).async {
     implicit request =>
-      connector.getReturn(periodKey, vpdId = request.enrolmentVpdId)
-        .map { returnData =>
-          Ok(view(ViewIndividualReturnViewModel(returnData)))
-        }
+      for {
+        returnData <- connector.getReturn(periodKey, request.enrolmentVpdId)
+        dutyRateOpt <- obligationService.getDutyRateForPeriod(request.enrolmentVpdId, periodKey)
```
ian-hmrc: We should not look up the duty rate by date here, we should play back the duty rate that comes in the return (in particular, it can be found down in `RegularReturn.dutyRate`).

The question remains as to what to do when there is a nil return. Maybe the _'Duty is calculated by multiplying the volume of vaping products released for consumption by the Vaping Product Duty rate of {0} per 10ml.'_ content needs to change for this scenario as it does not seem appropriate?

app/controllers/returns/submit/CheckYourAnswersController.scala
ian-hmrc: It seems we're only supplying a duty rate for the _Duty is calculated by multiplying the volume of vaping products released for consumption by the Vaping Product Duty rate of {0} per 10ml._ message. When a nil return is being submitted, this message seems inappropriate. We should check that scenario with the content designer. This may leave us with only having to do this for the non-nil return case. With the addition of the task-based CYA pages, coming soon, maybe we should store the duty rate at that point and then just use that to play-back here?

test/controllers/returns/submit/CheckYourAnswersControllerSpec.scala
```diff
@@ -36,16 +36,46 @@ class CheckYourAnswersControllerSpec extends SpecBase {
...
+
+        val view = application.injector.instanceOf[CheckYourAnswersView]
+        val vm = CheckYourAnswersViewModel(returnsUserAnswers, testDutyRate)(messages(application))
+
+        status(result) mustEqual OK
+        contentAsString(result) mustEqual view(vm)(request, messages(application)).toString
+      }
+    }
+
+    "must use zero duty rate when obligation service returns None" in {
```
ian-hmrc, line 61: I think we need to question this assumption as this is not part of the specification. I've explained this above in the Controller comments.

app/controllers/returns/submit/CheckYourAnswersController.scala
```diff
@@ -35,18 +36,27 @@ class CheckYourAnswersController @Inject()(
...
+      val vm = CheckYourAnswersViewModel(request.userAnswers, dutyRate)
+      Ok(view(vm))
+    }.recover { _ =>
+      Redirect(controllers.routes.JourneyRecoveryController.onPageLoad())
```
ian-hmrc: Why go to Journey Recovery here instead of just showing the error page? In this situation, either the duty rates are mis-configured, or there isn't a duty rate for the given date (which is again a system mis-configuration). The user cannot recover the journey from this. I'd be happy for this to just be dealt with by the error page rather than trying to be too clever.

test/controllers/returns/view/ViewIndividualReturnControllerSpec.scala
```diff
@@ -19,39 +19,80 @@ package controllers.returns.view
...
+        val vm = ViewIndividualReturnViewModel(createReturnDisplayResponse(), testDutyRate)
+
+        val view = application.injector.instanceOf[ViewIndividualReturnView]
+
+        status(result) mustEqual OK
+        contentAsString(result) mustEqual view(vm)(request, messages(application)).toString
+      }
+    }
+
+    "must use zero duty rate when obligation service returns None" in {
```
ian-hmrc: Again, I think this needs checking with UCD and what we should do for Nil returns.

test/services/returns/ObligationServiceSpec.scala
```diff
@@ -0,0 +1,180 @@
...
+        
+        val service = new ObligationService(mockObligationsConnector, mockDutyRateService)
+        
+        val result = service.getDutyRateForPeriod(vpdId, "26XX").futureValue
+        
+        result mustBe None
+      }
+
+      "must return None when obligations response is empty" in {
+        val emptyResponse = ObligationsResponse(obligation = Seq.empty)
```
ian-hmrc: We seem to be holding onto the language of the connector here with 'response' here in `emptyResponse`. I think it would be much better to step away from this and use more domain-centric language, like `noObligations`. I think this approach could be considered elsewhere in the app and test code too.

I think that behind this is the fact that we've named our API data classes to closely express the HTTP domain, i.e. `ObligationsResponse`. However, they are also used throughout the app. On other projects in the past, I would have translated the response format into a domain format to help ensure against changes in the API not needing changes cascading everywhere, and to hide some of the unnecessary details (e.g. that `ObligationResponse` has a sequence of `ObligationItems` [named `obligation`!] each of which contains an `ObligationDetails` where the data lies with names like `iCToDate`). This could be translated to a simple, domain-oriented,  `Seq[Obligation]` where `Obligation` would have clearer fields like `from` and `to` (or possibly even a date range). I appreciate this would be quite a big change, and I'm not saying we should do it now, but it's worth considering the cost of exposing our API's data structures throughout the code base.

test/services/returns/SubmitReturnServiceSpec.scala
```diff
@@ -17,49 +17,151 @@
...
+    returnsUserAnswers
+  )
+
+  "SubmitReturnService" - {
+
+    "submit" - {
 
-  "SubmitReturnService must" - {
+      "must return success when a return is submitted successfully" in {
+        when(mockObligationService.getObligationByPeriodKey(eqTo(vpdId), eqTo(periodKey))(using any()))
```
ian-hmrc, line 69: This tests pretty much none of the complex business logic in the `SubmitReturnService`. It basically says that a submission is made, and it checks that the _stubbed_ `testReturnSubmissionResponse` comes back. It ignores all the calculation (other than the execution can pass through it, e.g. no null pointer exceptions).

I suspect that the Service needs decomposing into smaller functions (by task?... and then maybe smaller still???) and those should be unit tested directly. We'll be getting a high coverage value here which is lying to us about our real test thoroughness.

TBH, when the task-based CYA pages come back, some of these functions are likely to be distributed around the service, so having them free and mobile will be useful.

test/viewmodels/returns/CheckYourAnswersViewModelSpec.scala
```diff
@@ -17,49 +17,37 @@
...
 
-    "must filter out None values from summary rows" in {
-      val ua = ReturnsUserAnswers("id", periodKey, JsObject.empty, Instant.now(), Instant.now())
-        .set(DeclareDutyPage, true)
-        .flatMap(_.set(EnterDutyAmountPage, 1000)).success.value
+    "must create view model with correct summary lists" in {
+      val userAnswers = returnsUserAnswers.set(EnterDutyAmountPage, 500).success.value
 
-      val vm = CheckYourAnswersViewModel(ua)
+      val result = CheckYourAnswersViewModel(userAnswers, testDutyRate)
```
ian-hmrc: I'm just curious as to why the name of this val has changed from `vm` to `result`? The test name is 'must create _view model_ with correct summary lists'. Why not keep with the language being used?

app/config/DutyRateConfig.scala
ian-hmrc: Yes. It starts cleanly and only shows the error when submitting a return. I'd like the app to fail to start so that when a config change that fails validation is made, the person making that change gets that feedback immediately.

test/viewmodels/returns/CheckYourAnswersViewModelSpec.scala
craigemmerson: The assertion in the test is saying our result of x should match y. its used in most of our tests to call the assertion "result" so seemed like this was actually more in keeping with the rest of the app, no?

test/config/DutyRateConfigSpec.scala
craigemmerson: This spec will be removed after the created validation spec

app/config/DutyRateConfig.scala
craigemmerson: 👍

test/config/DutyRateConfigSpec.scala
craigemmerson: 👍

test/config/DutyRateConfigSpec.scala
craigemmerson: 👍

app/controllers/returns/view/ViewIndividualReturnController.scala
craigemmerson: now gathering rate from regular returns, but need to clarify designs with UCD

test/services/returns/SubmitReturnServiceSpec.scala
craigemmerson: Will create a follow up ticket to address test refinement

test/services/returns/ObligationServiceSpec.scala
ian-hmrc: Happy with the rename for now. We can think about the coupling to the API repesentation in the medium to long term.

test/viewmodels/returns/CheckYourAnswersViewModelSpec.scala
ian-hmrc: I can see the rename which is good for now. 👍

We should be careful about our levels of abstraction going forward. `result` is too abstract as it could be used across all tests in the service. There is more value in maintaining the scope locally in each test, and expressing what exactly is being tested clearly.

app/config/DutyRateConfig.scala
ian-hmrc: Cool. The changes make this much clearer.

test/config/DutyRateConfigSpec.scala
ian-hmrc: I'm only suggesting the `val config` is inlined into the one test where it is being used here. The spec will need to be retained to test whatever behaviour is left in `DutyRateConfig`.

test/config/DutyRateConfigSpec.scala
ian-hmrc: I can see that has been broken down and is now more modular. :)

app/services/returns/ObligationService.scala
ian-hmrc: The approach that has been taken with the case classes like `DutyRateToPence` is not quite what I had in mind. The types being used here are still BigDecimals, and it is possible to put a ml value in a litres val. I was expecting something like `opaque type VpdId = String`, but to wrap our BigDecimal (though a case class for Litres, Millilitres etc. would also do).

app/controllers/returns/view/ViewIndividualReturnController.scala
ian-hmrc: I've asked for a meeting with CD tomorrow after standup.

app/models/returns/DutyRate.scala
ghousedev: We have introduced a DateRange class, please take a look.

app/models/returns/DateRange.scala
```diff
@@ -0,0 +1,38 @@
...
+final case class DateRange(start: LocalDate, end: LocalDate) {
+  
+  def contains(date: LocalDate): Boolean =
+    !date.isBefore(start) && !date.isAfter(end)
+  
+  def overlaps(other: DateRange): Boolean =
+    !this.end.isBefore(other.start) && !other.end.isBefore(this.start)
+  
+  def durationInDays: Long =
+    ChronoUnit.DAYS.between(start, end)
```
ian-hmrc: This method is untested and unused in the service. Moreover, I think it has an error because the range is inclusive of the first argument and exclusive of the end argument. Was this LLM generated? It seems superfluous to requirements.

app/config/DutyRateConfig.scala
```diff
@@ -0,0 +1,50 @@
...
+@Singleton
+class DutyRateConfig @Inject()(configuration: Configuration) {
+  
+  private val DUTY_RATES_KEY = "duty-rates"
+  
+  val rates: Seq[DutyRate] = {
+    val configList = configuration.get[Seq[Configuration]](DUTY_RATES_KEY)
+    val parsedRates = configList.map { rateConfig =>
+      DutyRate(
+        period = models.returns.DateRange(
```
ian-hmrc: Would it be possible to import `DateRange` here to reduce the noise in this code?

app/models/returns/DateRange.scala
ghousedev: I am planning to remove those unused methods as the Validator has the behaviour we need.

app/models/returns/DutyRate.scala
ian-hmrc: I can see that commit removed the `DutyRateConfigSpec`. I still think that requires a testing. I suggest that it might be simpler to decompose the long initialisation block in there into a parsing function and a validation function and test those more directly?

test/config/DutyRateValidatorSpec.scala
```diff
@@ -0,0 +1,358 @@
...
+
+  "validateNonEmpty" - {
+
+    "must return Right for a non-empty sequence" in {
+      val rates = Seq(validRate1)
+
+      val result = DutyRateValidator.validateNonEmpty(rates)
+
+      result mustBe Right(rates)
+    }
```
ian-hmrc, line 53: Just as a point of style, I find tests often easier to read when not trying so hard to follow the given/when/then pattern, for example:

```
    "must return Right for a non-empty sequence" in {
      validateNonEmpty(Seq(validRate1)) mustBe Right(Seq(validRate1))
    }
```

app/config/DutyRateConfig.scala
ian-hmrc: Cool :)

app/controllers/returns/submit/CheckYourAnswersController.scala
ghousedev: There is no generic error controller in the service at the moment.  Unless I am missing something?

ian-hmrc: The `ErrorHandler` shows the `ErrorTemplate` View, no? https://github.com/hmrc/vaping-duty-frontend/blob/main/app/handlers/ErrorHandler.scala

ghousedev: We would need some content to be provided here, unless there is some default title and message we want to display?

ghousedev: Removing the recover block has directed to the error template

app/controllers/returns/submit/CheckYourAnswersController.scala
ian-hmrc: Cool. We've raised a ticket with content for this and I can see that we now put out placeholder message for the nil return case. We can pick this up when a decision is made about what to show.

app/models/returns/DutyRate.scala
ian-hmrc: Cool. It's good to see the test come back, but I still think we should break it down a little and test more locally. I see there could be a parsing function that would make a nice unit, and another for reacting to the validation failure by throwing the exception (it would be nice to inject the validator so that it could be replaced by a mock in the tests). If this is a bit much to get done by the end of the sprint then that's fine and we can pick this up as a bit of tech debt another time.

app/services/returns/ObligationService.scala
ian-hmrc: OK. I've had a further look through more widely the code on the back of this. It seems that the calculation is off because of some incorrect conversions. It also seems like some of the calculations are duplicated for the CYA page and the submission. I think we should look at de-duplicating that and making a single source of truth for these calculated elements. I'd also like to see them broken down a bit more, for example the basic duty, spoilt, under/over calculations should be in separate functions which can be unit tested directly. I'd suggest that we just get the calculation correct for now, and maybe we can pick this up as another piece of work later on. I'd like to be in the mob for that if possible.

app/controllers/returns/view/ViewIndividualReturnController.scala
ian-hmrc: OK, that meeting didn't happen and now the CD is on holiday for a week. We have a ticket to cover this though, so let's consider this one done for the purposes of this PR.

test/controllers/returns/view/ViewIndividualReturnControllerSpec.scala
ian-hmrc: This will be covered by the ticket we've made for UCD. Happy to close this for the purposes of this PR.

app/models/returns/DutyRate.scala
ian-hmrc: I can see this is done :)

conf/application.conf
ghousedev: Have created a PR for app-config-base and added some comments in the both the application and base.

app/services/returns/ObligationService.scala
ian-hmrc: Cool. :)

app/config/DutyRateConfig.scala
ian-hmrc: We can check this on Staging with a broken config once it has been merged.

test/controllers/returns/submit/CheckYourAnswersControllerSpec.scala
ian-hmrc: OK, this will be picked up on the back of the UCD ticket we've spun off. Happy for this to go ahead as-is.

test/services/returns/SubmitReturnServiceSpec.scala
ian-hmrc: I've created [VPD-1775](https://jira.tools.tax.service.gov.uk/browse/VPD-1775) to address this.

ian-hmrc (approved):
LGTM

PR #97, opened 2026-06-03
VPD-1692: Hide adjustments section when no fulfilled obligations are …

ghousedev:
…present

### DIFF
--- a/app/models/returns/AdjustmentsEligibility.scala
+++ b/app/models/returns/AdjustmentsEligibility.scala
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2026 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


test/models/returns/AdjustmentsEligibilitySpec.scala
```diff
@@ -0,0 +1,92 @@
...
+        createObligation(ObligationStatus.O),
+        createObligation(ObligationStatus.O),
+        createObligation(ObligationStatus.O)
+      )
+
+      AdjustmentsEligibility.fromObligations(obligations) mustBe AdjustmentsEligibility.NotEligible
+    }
+  }
+
+  "AdjustmentsEligibility.isEligible" - {
```
craigemmerson: I think these two tests add no value , we just assert the hard coded values in the models are what we set. the from obligations tests are where the logic is covered. I think we can remove these two

app/models/returns/AdjustmentsEligibility.scala
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.returns
+
+import models.obligations.{ObligationItem, ObligationStatus}
+
+sealed trait AdjustmentsEligibility {
+  def isEligible: Boolean
```
craigemmerson: if we do end up removing the tests, we can remove this function as its only ever used In tests

craigemmerson (approved):

PR #98, opened 2026-06-04
VPD-1663: Clear user-answers when submission of return is complete

ghousedev:


craigemmerson (dismissed):

app/controllers/returns/submit/CheckYourAnswersController.scala
```diff
@@ -46,15 +47,20 @@ class CheckYourAnswersController @Inject()(
       VpdId(request.userAnswers.vpdId),
       request.periodKey
     ).map { dutyRateOpt =>
-      val dutyRate = dutyRateOpt.getOrElse(BigDecimal(0))
+      val dutyRate = dutyRateOpt.getOrElse(BigDecimal("0"))
```
ian-hmrc: We should not be defaulting the duty rate to zero. I know this is unlikely, and we have safeguards to protect against this, but if this does happen, we have the possibility of charging the manufacturer at zero rate, which would be a serious issue for HMRC. Can we instead make this an error?

app/controllers/returns/submit/CheckYourAnswersController.scala
```diff
@@ -46,15 +47,20 @@ class CheckYourAnswersController @Inject()(
...
     }
   }
 
   def onSubmit: Action[AnyContent] = (identify andThen returnsEnabled andThen getData andThen requireData).async { implicit request =>
-    submitReturnService.submit(request.userAnswers).map { response =>
-      Redirect(controllers.returns.submit.routes.ConfirmationController.onPageLoad().url + s"?period=${request.periodKey}")
+    submitReturnService.submit(request.userAnswers).flatMap { response =>
+
+      userAnswersService.clear(request.enrolmentVpdId, request.periodKey).map {
+        case Left(_)  => Redirect(controllers.routes.JourneyRecoveryController.onPageLoad())
```
ian-hmrc: I'm not sure we should send the user to the Journey Recovery Controller here. How would they be able to recover the journey in this situation? I think this is more of an error case.

This has been making me think about what could justify going to the Journey Recovery Controller. I'm starting to think that if some data that the user should have entered is missing, and if they just re-run the journey they would fill it in, then going to the JRC and competing the missing parts of the journey to provide the missing data would be useful. This is different to un-recoverable situations like this.

Further to this, since the frontend supports several journeys, we may need a JRC for each journey. However, this is one to chat through really.

ghousedev: Had a quick discussion with Sam and if submission is a success then we should still show the confirmation page, the specific ua document for the id and period would sit in mongo until TTL expiration in the case of clearing failing.

May not be the best thing to do but it seems to make sense.

app/controllers/returns/submit/CheckYourAnswersController.scala
ghousedev: Implemented the same as in the DeclareDutyCheckAnswersController, we can look to remove duplication once one of the PRs has been merged and we rebase.

test/controllers/returns/submit/CheckYourAnswersControllerSpec.scala
```diff
@@ -58,7 +58,7 @@ class CheckYourAnswersControllerSpec extends SpecBase {
       }
     }
 
-    "must use zero duty rate when obligation service returns None" in {
+    "must throw RuntimeException when obligation service returns None" in {
```
ian-hmrc: Might I suggest we rename this in domain terms "must send the user to the error page if no duty rate is found for the month of the return"?

app/controllers/returns/submit/CheckYourAnswersController.scala
ian-hmrc: OK, I'm happy with that approach. However, I don't think the code does that for all scenarios. If the call to `userAnswersService.clear()` returns a failed future, then the user will still be sent to the journey recovery controller.

I think the longer term trick here is to return `Future[Unit]` from `clear()` instead of `Future[Either[UpstreamErrorResponse, Unit]]`. However, this might entail more work than we would like right now. To get this over the line, we could just add the missing test for the failed future, ensure it passes, and pick this refactoring up at a later date.

app/services/returns/TaskStatusService.scala
```diff
@@ -47,14 +47,18 @@ object TaskStatusService {
     }
   }
 
-  def allTasksCompleted(answers: ReturnsUserAnswers): Boolean = {
+  def allTasksCompleted(answers: ReturnsUserAnswers, adjustmentsEligibility: AdjustmentsEligibility): Boolean = {
+    val spoiltTaskComplete = adjustmentsEligibility match {
+      case AdjustmentsEligibility.NotEligible => true
+      case AdjustmentsEligibility.Eligible    => declareSpoiltProductsTaskStatus(answers) == TaskStatus.Completed
```
ian-hmrc: This seems like quite a complex way to achieve this outcome. We will need the Y/N question for the adjustments to be set to 'No' for the submission (eventually). Might there be a better way to achieve this by just setting those values earlier on (may from the Before You Start page)? This code would then not need changing, and nothing special would need to be done for the submission code.

app/services/returns/TaskListPreparationService.scala
```diff
@@ -0,0 +1,53 @@
...
+                                            userAnswers: ReturnsUserAnswers,
+                                            answer: Boolean
+                                          )(using HeaderCarrier): Future[ReturnsUserAnswers] = {
+
+    userAnswers.get(DeclareSpoiltProductsPage) match {
+      case Some(value) if value == answer => Future.successful(userAnswers)
+      case _ =>
+        userAnswers.set(DeclareSpoiltProductsPage, answer).fold(
+          _   => Future.successful(userAnswers),
+          ua  => repository.set(ua).map(_ => ua)
```
ian-hmrc: I think this code could be made simpler by separating the action of updating the UA for the first return from saving the updated UA to the repo. It would also make the tests easier to write.

app/services/returns/TaskStatusService.scala
ian-hmrc: Cool. I can see that is done from the task list page and the service has reverted to being simpler.

app/services/returns/TaskListPreparationService.scala
ian-hmrc: I've made that change.

craigemmerson (approved):

PR #102, opened 2026-06-05
VPD-1779: Mini CYA page for Declare duty task

ghousedev:


app/controllers/returns/submit/DeclareDutyCheckAnswersController.scala
```diff
@@ -0,0 +1,57 @@
...
+  val controllerComponents: MessagesControllerComponents,
+  view: DeclareDutyCheckAnswersView
+)(using ExecutionContext) extends FrontendBaseController with I18nSupport {
+
+  def onPageLoad: Action[AnyContent] = (identify andThen returnsEnabled andThen getData andThen requireData).async { implicit request =>
+    obligationService.getDutyRateForPeriod(
+      VpdId(request.userAnswers.vpdId),
+      request.periodKey
+    ).map { dutyRateOpt =>
+      val dutyRate = dutyRateOpt.getOrElse(BigDecimal(0))
```
ian-hmrc: It should be very unlikely that we don't have a duty rate, but could we just handle this by using `flatMap()`, maybe even as a for comprehension?

app/controllers/returns/submit/DeclareDutyCheckAnswersController.scala
```diff
@@ -0,0 +1,57 @@
...
+  returnsEnabled: ReturnsEnabledAction,
+  obligationService: ObligationService,
+  val controllerComponents: MessagesControllerComponents,
+  view: DeclareDutyCheckAnswersView
+)(using ExecutionContext) extends FrontendBaseController with I18nSupport {
+
+  def onPageLoad: Action[AnyContent] = (identify andThen returnsEnabled andThen getData andThen requireData).async { implicit request =>
+    obligationService.getDutyRateForPeriod(
+      VpdId(request.userAnswers.vpdId),
+      request.periodKey
```
ian-hmrc: Extracting vars for `vpdId` and `periodKey` simplifies this code a lot and allows this call to be all on one line.

app/controllers/returns/submit/DeclareDutyCheckAnswersController.scala
```diff
@@ -0,0 +1,57 @@
...
+  requireData: ReturnsDataRequiredAction,
+  returnsEnabled: ReturnsEnabledAction,
+  obligationService: ObligationService,
+  val controllerComponents: MessagesControllerComponents,
+  view: DeclareDutyCheckAnswersView
+)(using ExecutionContext) extends FrontendBaseController with I18nSupport {
+
+  def onPageLoad: Action[AnyContent] = (identify andThen returnsEnabled andThen getData andThen requireData).async { implicit request =>
+    obligationService.getDutyRateForPeriod(
+      VpdId(request.userAnswers.vpdId),
```
ian-hmrc: Is there a reason for using `request.userAnswers.vpdId` rather than `request.enrolmentVpdId`?

app/viewmodels/returns/submit/DeclareDutyCheckAnswersViewModel.scala
```diff
@@ -0,0 +1,84 @@
...
+  summaryList: SummaryList
+)
+
+object DeclareDutyCheckAnswersViewModel {
+
+  private val ZERO = "0"
+  private val ML_SUFFIX = " ml"
+
+  def apply(userAnswers: ReturnsUserAnswers, dutyRate: BigDecimal, periodKey: PeriodKey)(implicit messages: Messages): DeclareDutyCheckAnswersViewModel = {
+    val volumeInMl = userAnswers.get(EnterDutyAmountPage).getOrElse(0)
```
ian-hmrc: If the user has managed to get to the CYA page without a value, I think the page should fail to render rather than showing zero instead.

app/controllers/returns/submit/DeclareDutyCheckAnswersController.scala
ian-hmrc: I think the complexity here is not helped by the fact that `obligationService.getDutyRateForPeriod()` returns a `Future[Option[BigDecimal]]`. I think this represents 2 kinds of failure, first a failure to get the obligations from the API, and second, to find the obligation we're looking for in the successfully returned obligations. I think that as far as calling code is concerned, these are both simply failures to get the obligation and can be treated as the same. So I think we could do our trick of collapsing the return type to a `Future[BigDecimal]`, and that would make this code simpler.

app/controllers/returns/submit/DeclareDutyCheckAnswersController.scala
ghousedev: Using the id from the request removes the need to wrap the value so I have made that change.

app/controllers/returns/submit/DeclareDutyCheckAnswersController.scala
ghousedev: Moved to one line without extracting variables here, seems simple and readable enough to me but let me know.

app/controllers/returns/submit/DeclareDutyCheckAnswersController.scala
ghousedev: Have updated the controller

app/viewmodels/returns/submit/DeclareDutyCheckAnswersViewModel.scala
ghousedev: Would it be a good idea to handle this case once this page has the Yes/No value, as having no volume when a user answers 'No' will be legitimate at that point.

app/controllers/returns/submit/DeclareDutyCheckAnswersController.scala
```diff
@@ -0,0 +1,60 @@
...
+      val vm = DeclareDutyCheckAnswersViewModel(request.userAnswers, dutyRate, request.periodKey)
+      Ok(view(request.periodKey, vm))
+    }
+  }
+
+  def onSubmit: Action[AnyContent] = (identify andThen returnsEnabled andThen getData andThen requireData) { implicit request =>
+    Redirect(controllers.returns.submit.routes.TaskListController.onPageLoad().url + s"?period=${request.periodKey.value}")
+  }
+
+  private def getDutyRate(obligationService: ObligationService)(using request: ReturnsDataRequest[?]) = {
```
ian-hmrc: There is no need to pass the `obligationsService` here as it is a field of the class. I think this would be better parameterised with `vpdId` and `periodKey`, i.e. don't rely on the request to pass these but do it more explicitly.

app/viewmodels/returns/submit/DeclareDutyCheckAnswersViewModel.scala
ian-hmrc: We aren't currently showing the mini CYA page for the 'No' scenario. Even if we did, I don't think we should try to substitute in a default value for the volume. We don't need to submit anything on the API in this case it the cleanest model is just not to have this data in this scenario.

If it is they Yes case, and by some unlikely, contrived circumstances we have no volume, then again we should not be inventing a volume. I think this send the user to the Journey Recovery Controller and they should start again, so that they should add that missing data.

app/views/returns/submit/BeforeYouStartView.scala.html
```diff
@@ -38,10 +38,10 @@
 
     @govukWarningText(WarningText(
         iconFallbackText = Some(messages("site.warning")),
-        content = Text(messages("returns.beforeYouStart.deadline", vm.dueDate, vm.year.toString))
+        content = Text(messages("returns.beforeYouStart.deadline", vm.dueDate, vm.dueYear.toString))
```
ian-hmrc, line 41: We should also pass the day of the month from the API rather than assuming it will always be 7.

ghousedev: Could we get a ticket in for next sprint to handle this change?  Or even to bring in to the current, I just feel the scope of this ticket has expanded quite a lot as is.

ian-hmrc: I feel this isn't a big change. Any chance we can just get this done? I think the overhead of another ticket would be more than the work.

app/viewmodels/returns/submit/DeclareDutyCheckAnswersViewModel.scala
ian-hmrc: Cool. I can see that has been implemented. :)

app/views/returns/submit/BeforeYouStartView.scala.html
ian-hmrc: Let's pick this up in another PR.

ian-hmrc (dismissed):
LGTM :)

ian-hmrc (approved):
LGTM :)

PR #103, opened 2026-06-09
Vpd 1745

craigemmerson:


app/controllers/returns/submit/DeclarationController.scala
```diff
@@ -0,0 +1,79 @@
...
+    implicit request =>
+      form.bindFromRequest().fold(
+        formWithErrors =>
+          Future.successful(BadRequest(view(request.periodKey, formWithErrors))),
+
+        value =>
+          for {
+            updatedAnswers <- Future.fromTry(request.userAnswers.set(DeclarationPage, value))
+            _              <- sessionRepository.set(updatedAnswers)
+            _              <- submitReturnService.submit(updatedAnswers)
```
ghousedev: Could we handle a failed submission here?  In the case of a failed submission I don't think we want to clear a users inputs from user answers.

app/models/returns/submit/ReturnCreateRequest.scala
```diff
@@ -16,15 +16,16 @@
...
+import models.returns.{DeclarationDetails, TotalDutyDue, VapingProductsProduced}
 import play.api.libs.json.{Json, OFormat}
 
-case class ReturnCreateRequest(
+final case class ReturnCreateRequest(
   periodKey: String,
   vapingProductsProduced: VapingProductsProduced,
-  totalDutyDue: TotalDutyDue
+  totalDutyDue: TotalDutyDue,
+  declaration: DeclarationDetails
```
ghousedev, line 26: I think we will also need to add this field to the ReturnDisplaySuccess model to use when playing back returns to users.

ghousedev (approved):
👍

PR #104, opened 2026-06-11
VPD-1780: Mini CYA page for Duty Suspension journey

ghousedev:


app/viewmodels/checkAnswers/DutySuspenseCheckAnswersSummary.scala
```diff
@@ -0,0 +1,66 @@
...
+    SummaryListRowViewModel(
+      key = KeyViewModel(HtmlContent(keyHtml)),
+      value = ValueViewModel(HtmlContent(valueHtml)),
+      actions = Seq(
+        ActionItemViewModel(
+          "site.change",
+          s"${controllers.returns.submit.routes.EnterDutySuspenseController.onPageLoad(NormalMode).url}?period=${periodKey.value}"
+        ).withVisuallyHiddenText(messages("returns.dutySuspenseCheckAnswers.change.hidden"))
+      )
+    )
```
ian-hmrc, line 63: Can I ask if you queried the design here? Seems like we're faking 2 rows in this table by using line breaks. It doesn't feel like it will fail accessibility.

ian-hmrc: Actually, I've just spoken to the designer and he's pointed out that a later rev will fix this. I'm happy for this to go as-is for now as we'll be making this change quite soon.

ian-hmrc (dismissed):
LGTM :)

ian-hmrc (dismissed):
LGTM :)

ian-hmrc (approved):
LGTM :)

PR #106, opened 2026-06-18
Vpd 1779 refactor tests

ian-hmrc:
Introduce easier building of obligations in tests

Work from BeforeYouStatViewModelSpec making the tests more direct and introducing obligation builders to reduce the amount of code in the tests and be more flexible. Fixed up a few issues that became evident along the way.

This has only been applied in 1 test, and could be rolled out over more tests given time.


ghousedev (approved):
👍

PR #107, opened 2026-06-18
Vpd 1812 - Declare vaping products duty new content

craigemmerson:


app/services/returns/SubmitReturnService.scala
```diff
@@ -56,7 +56,7 @@ class SubmitReturnService @Inject()(
 
     val zeroValue = BigDecimal("0")
     val dutyDeclared = ua.get(DeclareDutyPage).getOrElse(false)
-    val liquidInMl = ua.get(EnterDutyAmountPage).fold(zeroValue)(value => BigDecimal(value))
+    val liquidInMl = ua.get(EnterDutyAmountPage).getOrElse(zeroValue)
```
ghousedev, line 59: We should use validation on the cya page to ensure we cannot submit a zero accidentally, not for this PR but something we should do.

ghousedev (approved):
👍

PR #110, opened 2026-06-22
Vpd 1810 - Main check your answers for day 1

craigemmerson:


app/models/returns/ReturnsUserAnswers.scala
```diff
@@ -29,6 +29,8 @@ import scala.util.{Failure, Success, Try}
 final case class ReturnsUserAnswers(
                                      vpdId: String,
                                      periodKey: String,
+                                     returnPeriod: Option[String] = None,
```
ghousedev: Could we store this as a Month, so we can using the month directly in the ReturnsDateUtils.getReturnMonth() function.

ghousedev (approved):
👍

PR #111, opened 2026-06-22
VPD-1719: Update View multiple returns to latest version

ghousedev:


app/controllers/returns/view/ViewMultipleReturnsController.scala
```diff
@@ -19,32 +19,56 @@ package controllers.returns.view
...
                                                override val messagesApi: MessagesApi,
+                                               clock: Clock,
                                                identify: ApprovedVapingManufacturerAuthAction,
                                                returnsEnabledAction: ReturnsEnabledAction,
                                                obligationService: ObligationService,
                                                val controllerComponents: MessagesControllerComponents,
                                                getData: DataRetrievalAction,
                                                view: ViewMultipleReturnsView
-                                             )(using ExecutionContext) extends FrontendBaseController with I18nSupport {
+                                             )(using ExecutionContext, Clock) extends FrontendBaseController with I18nSupport {
```
ian-hmrc: We're passing the `Clock` here implicitly to be able to get `LocalDate.now()` in a bunch of places down in the call stack, as well as injecting an instance of the Clock as the parameter `clock`. Have you considered just making `now` a parameter and passing it from the Controller where necessary? This would make the functions pure and make writing the tests simpler.

app/viewmodels/returns/view/ViewMultipleReturnsViewModel.scala
```diff
@@ -37,25 +42,63 @@ object ViewMultipleReturnsViewModel {
...
+          number = Some(year.toString),
+          href = controllers.returns.view.routes.ViewMultipleReturnsController.onPageLoad(Some(year)).url,
+          current = Some(year == currentYear)
+        )
+      }
+      Some(PaginationViewModel(paginationItems))
+    } else None
+
+    val completedSections = Seq(
+      CompletedReturnsSection(
```
ian-hmrc: Why do we have a `Seq[CompletedReturnSection]` here when we only ever have on `CompletedReturnSection`? The years are iterated over within the `CompletedReturnsSection`. It looks to me that the outer `Seq` is redundant. Is this Claude getting carried away?

app/viewmodels/returns/view/ViewMultipleReturnsViewModel.scala
```diff
@@ -37,25 +42,63 @@ object ViewMultipleReturnsViewModel {
   private val TAG_CLASS_BLUE = "govuk-tag--blue"
   private val TAG_CLASS_RED = "govuk-tag--red"
 
-  def apply(obligationsResponse: ObligationsResponse)(implicit messages: Messages): ViewMultipleReturnsViewModel = {
+  def apply(obligationsResponse: ObligationsResponse, currentYear: Int)
+           (implicit messages: Messages, clock: Clock): ViewMultipleReturnsViewModel = {
 
-    val outstandingObligations = obligationsResponse.obligation
+    val outstandingItems = obligationsResponse.obligation
       .filter(_.obligationDetails.openOrFulfilledStatus == STATUS_OPEN.toString)
```
ian-hmrc: Have you considered using `partition` to deduplicate this filtering for outstanding/completed obligations? It also stresses that there are only 2 options here (though this could be enhanced by using types rather than strings).

```
val (outstandingObligations, fulfilledObligations) =
  obligationsResponse.obligation.partition(_.obligationDetails.openOrFulfilledStatus == STATUS_OPEN.toString)

```

app/viewmodels/returns/view/ViewMultipleReturnsViewModel.scala
ghousedev: No, this sequence is used in the view, each year has it's own section.

ian-hmrc: I agree that there is one `CompletedReturnsSection` per year. However, since we only show one year at a time on the page, there is only ever one `CompletedReturnSection` in the `Seq`. By removing the outer `Seq`, we can remove the `@for()` in the view...

```
    @for(section <- vm.completedReturnsSections) {
        @h2(messages("returns.overview.completed.heading", section.year))

        @if(section.showEmptyMessage) {
            @p(Seq(messages("returns.overview.completed.empty")))
        } else {
            @govukTaskList(TaskList(items = section.items))
        }
    }
```

BTW, I've just noticed that `currentYear` is a bit misleading as a name. It's the year on the query parameter (or a default). Maybe it would be better named just as `year`.

app/controllers/returns/view/ViewMultipleReturnsController.scala
ghousedev: Have made some changes here

app/viewmodels/returns/view/ViewMultipleReturnsViewModel.scala
ian-hmrc: Cool. I can see that 's been done.

ian-hmrc (approved):
LGTM :)

PR #112, opened 2026-06-23
Vpd 1603 - Build view payments

craigemmerson:


app/views/payments/ViewPaymentsView.scala.html
```diff
@@ -0,0 +1,58 @@
...
+)
+
+@(vm: ViewPaymentsViewModel)(implicit request: Request[_], messages: Messages)
+
+@layout(pageTitle = titleNoForm(messages("payments.viewPayments.title"))) {
+
+    @h1(messages("payments.viewPayments.heading"))
+
+    <h1 class="govuk-heading-xl">
+        <span class="govuk-caption-l">@messages("payments.viewPayments.youOwe")</span>
```
ghousedev: Captions have a component that uses H2, which was pointed out in the last audit that we should do over using span.

app/connectors/payments/FinancialDataConnector.scala
```diff
@@ -0,0 +1,70 @@
...
+      .get(url"${config.getOutstandingPaymentsUrl(vpdId)}")
+      .execute[Either[UpstreamErrorResponse, HttpResponse]]
+      .recoverWith { case e: Exception =>
+        logger.warn(s"Exception while getting outstanding payments: ${e.getMessage}")
+        Future.failed(InternalServerException("Failed to get outstanding payments"))
+      }
+      .flatMap(getResponse)
+      .flatMap(parseJson)
+
+  private def getResponse(response: Either[UpstreamErrorResponse, HttpResponse]): Future[HttpResponse] = {
```
ian-hmrc: Might this method be named better? Maybe `logHttpErrors`? It would then also sit nicely alongside the `parseJson` on the next line.

app/controllers/payments/ViewPaymentsController.scala
```diff
@@ -0,0 +1,53 @@
...
+  service: FinancialDataService,
+  val controllerComponents: MessagesControllerComponents,
+  view: ViewPaymentsView
+)(using ExecutionContext) extends FrontendBaseController with I18nSupport with Logging {
+
+  def onPageLoad: Action[AnyContent] = identify.async { implicit request =>
+    service.getOutstandingPayments(request.enrolmentVpdId)
+      .map { payments =>
+        val paymentToDisplay = payments.headOption
+        val viewModel = ViewPaymentsViewModel(paymentToDisplay)
```
ian-hmrc: I know we're only going to show one payment for this ticket, but maybe we could just pass over the `Seq[OutstandingPayment]` rather than the `Option[OutstandingPayment]` as we're certain to need more of these very soon.

test/controllers/payments/ViewPaymentsControllerSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+          .thenReturn(Future.successful(Seq(testPayment)))
+
+        val application = applicationBuilder()
+          .overrides(bind[FinancialDataService].toInstance(mockService))
+          .build()
+
+        running(application) {
+          val request = FakeRequest(GET, routes.ViewPaymentsController.onPageLoad().url)
+          val result = route(application, request).value
+          val view = application.injector.instanceOf[ViewPaymentsView]
```
ian-hmrc: The IDE is complaining that `view` is unused. Judging from the test name, there is an assertion missing.

app/views/payments/ViewPaymentsView.scala.html
```diff
@@ -0,0 +1,56 @@
...
+            HeadCell(content = Text(messages("payments.viewPayments.table.description"))),
+            HeadCell(content = Text(messages("payments.viewPayments.table.leftToPay")), classes = "govuk-table__header--numeric"),
+            HeadCell(content = Text(messages("payments.viewPayments.table.status"))),
+            HeadCell(content = Text(messages("payments.viewPayments.table.action")))
+        )),
+        rows = vm.paymentRows,
+        classes = "govuk-table"
+    ))
+
+    @if(vm.payment.isEmpty) {
```
ian-hmrc: We could just check `vm.paymentRows` for being empty here and save having to add `payment` to the view model.

app/connectors/payments/FinancialDataConnector.scala
craigemmerson: Do you mean just rename to reflect the actions more? or actually combine with the parseJson function? 
"sit nicely alongside the parseJson" suggests you want them to combine to one operation.

app/controllers/payments/ViewPaymentsController.scala
craigemmerson: I'll have a look to change this, the data we get back from the backend work returns multiple payments so far so this seemed the cleanest way to only satisfy single payment for this ticket. I agree it will change soon anyway. Possibly for now move the headOption to the displaying side of this journey then for now?

test/controllers/payments/ViewPaymentsControllerSpec.scala
craigemmerson: ahh yes, this test was asserting things like 
contentAsString(result) must include("Vaping Products Duty payments")
but I decided to keep inline with the rest of the service we shouldn't be checking content so I removed the assertion. I will remove the view injection also

app/connectors/payments/FinancialDataConnector.scala
ian-hmrc: Sorry, that might have been unclear. I just meant rename `getResponse` to `logHttpErrors`. That would give the following code that sits together better...

```
      .flatMap(logHttpErrors)
      .flatMap(parseJson)
```

app/controllers/payments/ViewPaymentsController.scala
ian-hmrc: I think I'd rather it work with a `Seq`. That's where were going to be. If you're interested, I've got a commit sitting around for the change locally. I'd be happy to show you.

test/controllers/payments/ViewPaymentsControllerSpec.scala
ian-hmrc: OK, cool. May maybe then update the test name to reflect this? It's currently `"must return OK and the correct view when payments exist"`.

app/views/payments/ViewPaymentsView.scala.html
ian-hmrc: Cool. I can see that's been done.

ian-hmrc (approved):
LGTM :)

PR #113, opened 2026-06-25
VPD-1819: Update returns API to lates version

ghousedev:


app/models/returns/TotalDutyDue.scala
```diff
@@ -24,9 +24,9 @@ case class TotalDutyDue(
   totalDutyUnderDeclaration: BigDecimal,
   totalDutySpoiltProduct: BigDecimal,
   adjustmentAmount: BigDecimal,
```
ian-hmrc: `adjustmentAmount` is removed in v0.1.6

app/services/returns/SubmitReturnService.scala
```diff
@@ -67,40 +72,117 @@ class SubmitReturnService @Inject()(
...
+    val spoiltVolumes = ua.get(SpoiltVolumeByPeriodPage)
+    
+    spoiltVolumes match {
+      case Some(volumes) if volumes.nonEmpty =>
+        val dutyRateInPoundsPer10Ml = (BigDecimal(dutyRateInPencePerMl) / 100) * 10
+        
+        val spoiltProducts = volumes.map { spoiltVolume =>
+          val volumeInMl = BigDecimal(spoiltVolume.volume)
+          val volumeInLitres = ConvertToLitres(volumeInMl).toLitres
+          val dutyDue = (volumeInMl * (BigDecimal(dutyRateInPencePerMl) / 100)).setScale(2, BigDecimal.RoundingMode.DOWN)
```
ian-hmrc, line 143: This duty calculation is repeated around the code a few times now. We have a ticket coming to refactor the calculations. We can do that then.

app/services/returns/SubmitReturnService.scala
```diff
@@ -67,40 +72,117 @@ class SubmitReturnService @Inject()(
...
+      reasonForUnderDecl = None,
+      underDeclarationProducts = None
+    ))
+
+  private def buildOverDeclaration(): Option[OverDeclaration] =
+    Some(OverDeclaration(
+      overDeclFilled = FLAG_NOT_FILLED,
+      reasonForOverDecl = None,
+      overDeclarationProducts = None
+    ))
```
ian-hmrc, line 131: It's good to see default responses for the under/over going in. :)

app/services/returns/SubmitReturnService.scala
```diff
@@ -67,40 +72,117 @@ class SubmitReturnService @Inject()(
...
+      case _ =>
+        Some(SpoiltProduct(
+          spoiltProductFilled = FLAG_NOT_FILLED,
+          spoiltProducts = None
+        ))
+    }
+  }
+
+  private def buildOtherOptions(ua: ReturnsUserAnswers): Option[OtherOptions] = {
+    val dutySuspenseDeclared = ua.get(DeclareDutySuspensePage).getOrElse(false)
```
ian-hmrc, line 167: I think we should not use default values here. I know we should not be able to get to making the submission with this data incorrect, but we should fail to make the submission rather than assume we can just sweep this under the carpet.

We should probably also add the validation to the CYA page as a belt 'n' braces approach here.

PR #114, opened 2026-06-25
Vpd 1700 Cover all scenarios for View Individual Returns

craigemmerson:


app/controllers/returns/view/ViewIndividualReturnController.scala
```diff
@@ -43,18 +43,16 @@ class ViewIndividualReturnController @Inject()(
...
-          case Some(dutyRate) =>
-            Ok(view(ViewIndividualReturnViewModel(returnData, dutyRate)))
-          case None =>
-            logger.warn(s"No regular return found for period $periodKey")
-            Redirect(controllers.routes.JourneyRecoveryController.onPageLoad())
+        val dutyRate = extractDutyRate(returnData)
+        if (dutyRate.isEmpty) {
+          logger.warn(s"No regular return found for period $periodKey")
         }
+        Ok(view(ViewIndividualReturnViewModel(returnData, dutyRate)))
```
ian-hmrc: It feels like we're doing work here to extract the `dutyRate` from the `returnData` and then building the view model from the `returnData` and the `dutyRate`. Could we not just move this work inside the building of the view model?

app/controllers/returns/view/ViewIndividualReturnController.scala
```diff
@@ -43,18 +43,16 @@ class ViewIndividualReturnController @Inject()(
...
       connector.getReturn(periodKey, request.enrolmentVpdId).map { returnData =>
-        extractDutyRate(returnData) match {
-          case Some(dutyRate) =>
-            Ok(view(ViewIndividualReturnViewModel(returnData, dutyRate)))
-          case None =>
-            logger.warn(s"No regular return found for period $periodKey")
-            Redirect(controllers.routes.JourneyRecoveryController.onPageLoad())
+        val dutyRate = extractDutyRate(returnData)
+        if (dutyRate.isEmpty) {
+          logger.warn(s"No regular return found for period $periodKey")
```
ian-hmrc: I think this warning was added when we deliberately broke the view for nil returns. I don't think it needs to live on. Nil returns are a legitimate scenario and don't merit warnings being logged.

app/viewmodels/returns/view/ViewIndividualReturnViewModel.scala
```diff
@@ -181,10 +181,10 @@ case class ViewIndividualReturnViewModel(
     val totalRows = Seq(
       SummaryListRow(
         key = Key(
-          content = Text(messages("viewIndividualReturn.totalDutyDueVapingProducts"))
+          content = Text(messages("viewIndividualReturn.totalDutySpoiltProducts"))
```
ian-hmrc: I think I must have mis-directed you this morning when we spoke about this. On closer inspection, I think we want to keep the `totalDutyDueVapingProducts` as this should be at the bottom of the vaping duty section, before the spoilt section starts (that is not in the view yet). I imaging the spoilt total should be shown as part of the spoilt section once it arrives, but we need a design for that. We can do that when the spoilt section comes in to this view. I think the simplest action here is to drop this commit from the PR. Sorry for the confusion.

app/viewmodels/returns/view/ViewIndividualReturnViewModel.scala
```diff
@@ -32,12 +32,56 @@ case class ViewIndividualReturnViewModel(
...
+
+  private def formatPeriodKey(periodKey: String)(implicit messages: Messages): String = {
+    val year = 2000 + periodKey.take(2).toInt
+    val periodCode = periodKey.drop(2)
+
+    val monthNumber = (periodCode.charAt(1) - 'A') + 1
+    val month = java.time.Month.of(monthNumber)
+
+    s"${getMonthMessage(month)} $year"
+  }
```
ian-hmrc, line 121: Can we lookup the period key with the obligations API instead of assuming that this will work? I don't want us to be coupled to the format incase they change it. I know I did this in constructing test data, but I see that as a different case. We shouldn't do this in production code.

app/viewmodels/returns/view/ViewIndividualReturnViewModel.scala
```diff
@@ -89,82 +133,197 @@ case class ViewIndividualReturnViewModel(
       None
     }
 
-  def dutyTotalsSummaryList(implicit messages: Messages): SummaryList =
-    SummaryList(
-      rows = Seq(
-        SummaryListRow(
+  def dutyTotalsSummaryList(implicit messages: Messages): SummaryList = {
+    val spoiltProductsRows = spoiltProduct match {
+      case Some(sp) =>
```
ian-hmrc, line 187: A lot has been added to this function. My understanding is that it should build the rows for the final totals section which should be the last in this section. I think if we're going to add the spoilt, it should be done more like the `vapingProductsDeclarationSummaryList()` method, maybe as a `spoilProductsDeclarationSummaryList()` method?

ian-hmrc: OK, I can see the spoilt has been split out, but that method could probably do with renaming from `dutyTotalsSummaryList()` to `spoiltProductsSummaryList()` or similar.

ghousedev: Have renamed and refactored the viewmodel

ian-hmrc (approved):
LGTM :)

PR #117, opened 2026-06-26
VPD-1819b: Update returns API to latest version

ghousedev:


app/services/returns/SubmitReturnService.scala
```diff
@@ -67,40 +72,117 @@ class SubmitReturnService @Inject()(
...
+      case _ =>
+        Some(SpoiltProduct(
+          spoiltProductFilled = FLAG_NOT_FILLED,
+          spoiltProducts = None
+        ))
+    }
+  }
+
+  private def buildOtherOptions(ua: ReturnsUserAnswers): Option[OtherOptions] = {
+    val dutySuspenseDeclared = ua.get(DeclareDutySuspensePage).getOrElse(false)
```
ian-hmrc, line 167: We should not use default values here. I know we should not be able to get to making the submission with this data incorrect, but we should fail to make the submission rather than assume we can just sweep this under the carpet. You never know, the user might be able to get the data into a strange state by using the change links from the CYA page and then using the back button to back-out of a change part way through.

We should probably also add the validation to the CYA page as a belt 'n' braces approach here, but that can come in a future ticket.

app/services/returns/SubmitReturnService.scala
```diff
@@ -67,40 +72,117 @@ class SubmitReturnService @Inject()(
...
+            amountSpoilt = volumeInLitres,
+            dutyDue = dutyDue
+          )
+        }
+        
+        Some(SpoiltProduct(
+          spoiltProductFilled = FLAG_FILLED,
+          spoiltProducts = Some(spoiltProducts)
+        ))
+      case _ =>
```
ian-hmrc, line 158: Can we separate out the valid cases from the invalid cases here? If they respond 'No' but have `spoiltVolumes` or 'Yes' and no `spoiltVolumes` then they've managed to somehow may a mess of the data and we cannot assume that they just mean to not submit.

app/services/returns/TotalDutyDueCalculationService.scala
```diff
@@ -0,0 +1,67 @@
...
+    overDeclaration
+      .flatMap(_.overDeclarationProducts)
+      .map(_.map(_.dutyDue).sum)
+      .getOrElse(ZERO_VALUE)
+
+  private def calculateSpoiltProductTotal(spoiltProduct: Option[SpoiltProduct]): BigDecimal =
+    spoiltProduct
+      .flatMap(_.spoiltProducts)
+      .map(_.map(_.dutyDue).sum)
+      .getOrElse(ZERO_VALUE)
```
ian-hmrc, line 66: This feels like a lot of handling a lot of ways for something to be missing. We get a `Option[UnderDeclaration]` handed in where I would assume the None case is for No, yet the spoilt products inside is an `Option[Seq[SpoiltProductItem]]` it can also be, None, or an empty list. I feel that we might be able to collapse a bunch of these and simplify this code.

I'd have to see how it plays out when implemented, but I think we could pass an `UnderDeclaration` instead of an `Option[UnderDeclaration]`, and inside there the `spoiltProducts` could just be `Seq[SpoiltProductItem]`. We effectively use our validation to eliminate all the uncertainty and then can just work confidently with trusted data.

It's interesting to note that `SubmitReturnService.buildSpoiltProduct()` always returns `Some` and never `None`, so that needn't return an `Option`.

app/services/returns/SubmitReturnService.scala
ian-hmrc: OK, we've discussed this and will pick it up as part of the refactoring ticket to come.

app/services/returns/SubmitReturnService.scala
ian-hmrc: OK, we've discussed this and will pick it up as part of the refactoring ticket to come.

app/services/returns/TotalDutyDueCalculationService.scala
ian-hmrc: OK, we've discussed this and will pick it up as part of the refactoring ticket to come.

ian-hmrc (approved):
LGTM :)

PR #118, opened 2026-06-29
VPD 1855 submit return explicit audit event

ian-hmrc:
Sends an explicit audit event when submitting the VPD Return.


ghousedev (approved):
👍

PR #119, opened 2026-06-29
VPD-1856 Tighten validation to deny emails at top level domains

ian-hmrc:
Also add further unit tests to cover more cases


ghousedev (approved):
👍

PR #120, opened 2026-07-01
[VPD-1659] make volume forms follow correct validation

craigemmerson:


conf/messages.en
```diff
@@ -238,10 +238,10 @@ returns.spoiltVolumeByPeriod.heading = Tell us about your spoilt vaping products
 returns.spoiltVolumeByPeriod.caption = Spoilt vaping products — {0} {1}
 returns.spoiltVolumeByPeriod.label = How many millilitres do you need to declare as spoilt?
 returns.spoiltVolumeByPeriod.hint = Total millilitres of spoilt vaping products
-returns.spoiltVolumeByPeriod.error.required = You must enter a number greater than zero
-returns.spoiltVolumeByPeriod.error.wholeNumber = You must enter a number greater than zero
-returns.spoiltVolumeByPeriod.error.nonNumeric = You must enter a number greater than zero
-returns.spoiltVolumeByPeriod.error.outOfRange = You must enter a number greater than zero
+returns.spoiltVolumeByPeriod.error.required = Enter a volume with up to 12 digits and 1 decimal place, for example 999,999,999,999.9 ml
+returns.spoiltVolumeByPeriod.error.nonNumeric = Enter a volume with up to 12 digits and 1 decimal place, for example 999,999,999,999.9 ml
+returns.spoiltVolumeByPeriod.error.invalidDecimalPlaces = Enter 1 decimal place for amounts of 1 litre or more, for example 1.5 and 2 decimal places for amounts less than 1 litre, for example 0.75
```
ghousedev: Should the '1.5' not be something along the lines of '999,999,999,999.9' or '1000.5'

craigemmerson: I agree, I spoke a few times with Andrew and this was what the content document had. We should provide our own better message instead.

test/forms/returns/SpoiltVolumeByPeriodFormProviderSpec.scala
```diff
@@ -16,47 +16,75 @@
...
+        result.errors mustEqual Seq(FormError(fieldName, "returns.spoiltVolumeByPeriod.error.invalidDecimalPlaces"))
+      }
+    }
+
+    "must not bind values below the minimum of 1ml" in {
+      Seq("0.12", "0.99").foreach { input =>
+        val result = form.bind(Map(fieldName -> input)).apply(fieldName)
+        result.errors mustEqual Seq(FormError(fieldName, "returns.spoiltVolumeByPeriod.error.outOfRange", Seq(BigDecimal(1), BigDecimal("999999999999.9"))))
+      }
+    }
```
ian-hmrc, line 169: I can see these tests duplicated between all the FormProviderSpecs. Is there a way to test this logic more centrally, on the Formatter?

test/forms/returns/EnterDutySuspenseFormProviderSpec.scala
```diff
@@ -31,11 +31,26 @@ class EnterDutySuspenseFormProviderSpec extends FormSpec {
       val fieldName = "volumeReceived"
       val requiredKey = "returns.enterDutySuspense.volumeReceived.error.required"
       val nonNumericKey = "returns.enterDutySuspense.volumeReceived.error.nonNumeric"
+      val invalidDecimalKey = "returns.enterDutySuspense.volumeReceived.error.invalidDecimalPlaces"
       val outOfRangeKey = "returns.enterDutySuspense.volumeReceived.error.outOfRange"
```
ian-hmrc: `outOfRangeKey` is no longer used. Can it be removed?

test/forms/returns/EnterDutySuspenseFormProviderSpec.scala
```diff
@@ -64,11 +102,26 @@ class EnterDutySuspenseFormProviderSpec extends FormSpec {
       val fieldName = "volumeMoved"
       val requiredKey = "returns.enterDutySuspense.volumeMoved.error.required"
       val nonNumericKey = "returns.enterDutySuspense.volumeMoved.error.nonNumeric"
+      val invalidDecimalKey = "returns.enterDutySuspense.volumeMoved.error.invalidDecimalPlaces"
       val outOfRangeKey = "returns.enterDutySuspense.volumeMoved.error.outOfRange"
```
ian-hmrc: `outOfRangeKey` is no longer used. Can it be removed?

test/forms/returns/EnterDutySuspenseFormProviderSpec.scala
craigemmerson: have removed

test/forms/returns/EnterDutySuspenseFormProviderSpec.scala
craigemmerson: have removed

test/forms/returns/SpoiltVolumeByPeriodFormProviderSpec.scala
craigemmerson: I have tested inside mappings for the volume restriction instead to cover this

ian-hmrc: I still see the "must bind valid values >= 1000ml with no decimal places" and "must bind valid values < 1000ml with 0 or 1 decimal place" tests duplicated, etc. tests in all the FormProviderSpecs. Could this not be centralised?

test/forms/returns/EnterDutySuspenseFormProviderSpec.scala
ian-hmrc: This one is still present.

craigemmerson: ah you said outOfRangeKey which doesn't match the green line. I can remove the other too

test/forms/returns/SpoiltVolumeByPeriodFormProviderSpec.scala
craigemmerson: centralised where ? we could merge the forms into one as they are nearly identical, but that's breaking the pattern of all similar forms such as yes/no. we should keep forms separate so we see which form is for what page, and if we have separate forms we should have respective specs.

test/forms/returns/EnterDutySuspenseFormProviderSpec.scala
ian-hmrc: Ah, yes, sorry. Must have mis-clicked that line. I can see that's been done though.

test/forms/returns/SpoiltVolumeByPeriodFormProviderSpec.scala
ian-hmrc: OK, this seems harder that we'd like, given the framework. Let's leave this as it is for now.

app/services/returns/VolumePrecisionService.scala
```diff
@@ -0,0 +1,68 @@
...
+    val normalizedMaxVolume = normalizeMaxVolume(actualMaxVolume)
+
+    MaxVolumeResult(
+      maxVolumeInMl = normalizedMaxVolume,
+      formattedForDisplay = formatWithCommas(normalizedMaxVolume)
+    )
+  }
+
+  def normalizeMaxVolume(calculatedMax: BigDecimal): BigDecimal = {
+    if (calculatedMax <= 0) return calculatedMax
```
ian-hmrc: Scalafix is complaining about the early returns here when I run `sbt runAllChecks`.

craigemmerson: my bad, I only ran the tests not all checks. I have changed to simple if else to fix this

ian-hmrc (approved):
LGTM :)

PR #121, opened 2026-07-01
VPD-1775 initial refactoring

ian-hmrc:
Initial work to simplify the calculation in preparation to deal with inconsistent User Answers data.

### DIFF
--- a/test/services/returns/BuildReturnSubmissionServiceSpec.scala
+++ b/test/services/returns/BuildReturnSubmissionServiceSpec.scala
@@ -1,0 +1,225 @@
+/*
+ * Copyright 2026 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


test/services/returns/BuildReturnSubmissionServiceSpec.scala
```diff
@@ -0,0 +1,225 @@
...
+          .set(EnterDutySuspensePage, dutySuspenseVolumes).success.value
+          .set(DeclarationPage, declaration).success.value
+
+        val returnCreateRequest = service.buildSubmission(userAnswers, obligation, vpdId, Map(PeriodKey(obligation.periodKey) -> 1050))
+
+        returnCreateRequest.vapingProductsProduced.vapingProdManufactured mustBe yes
+        returnCreateRequest.vapingProductsProduced.returns.size mustBe 1
+        returnCreateRequest.vapingProductsProduced.returns.head mustBe RegularReturn("641", 105.0, 1, 10500)
+
+        //        returnCreateRequest.spoiltProduct.get.spoiltProductFilled mustBe no
```
ghousedev: DId you want to remove this comment?  Plus a few others in this file.

ian-hmrc: I was going to wait for the next PR for this one, but I've done that. It required setting-up more authentic user answers test data, so I've done that by starting from a completely nil return, and then allowed the tests to override the answers required for the particular scenario. That's allowed the code to be updated to test the spoilt in each test.

I feel like we're repeating assertions in the tests right now. I suspect these will get narrower as I add more cases for the broken user answers. Likely in the next PR.

ghousedev (dismissed):
👍

ghousedev (approved):
👍

PR #124, opened 2026-07-03
VPD-1396: Build initial adjustments pages

ghousedev:


app/controllers/returns/submit/adjustments/AdjustmentCheckYourAnswersController.scala
```diff
@@ -0,0 +1,144 @@
...
+      }
+  }
+
+  private def getDutyRatesForAdjustments(adjustmentList: Option[AdjustmentList])
+                                        (implicit request: ReturnsDataRequest[_]): Future[Map[String, BigDecimal]] = {
+
+    val adjustmentPeriods = adjustmentList.map(_.adjustments.map(_.period)).getOrElse(Seq.empty)
+
+    Future.sequence(adjustmentPeriods.map { period =>
+      dutyRateService.getDutyRate(request.enrolmentVpdId, period).map(rate => period.toString -> rate)
```
ian-hmrc: This will get the obligations once for each adjustment, which is inefficient. Can we do something to make this more efficient?

app/controllers/returns/submit/adjustments/AdjustmentCheckYourAnswersController.scala
```diff
@@ -0,0 +1,144 @@
...
+                }
+              },
+
+            value =>
+              for {
+                updatedAnswers <- Future.fromTry(request.userAnswers.set(AddAnotherAdjustmentPage, value))
+                _              <- sessionRepository.set(updatedAnswers)
+              } yield Redirect(navigator.nextPage(AddAnotherAdjustmentPage, NormalMode, updatedAnswers))
+          )
+      }
```
ian-hmrc: Quite a lot is being done in the Controller here, giving it several responsibilities above just building a view from the view model. For example, this controller pulls in both `ObligationService` and `DutyRateService`. Could some of this be extracted and pushed down?

app/controllers/returns/submit/adjustments/AdjustmentCheckYourAnswersController.scala
```diff
@@ -0,0 +1,144 @@
...
+    )
+  }
+
+  private def withObligations(
+    block: ObligationsResponse => Future[Result]
+  )(implicit request: ReturnsDataRequest[AnyContent]): Future[Result] = {
+    obligationService.getObligations(request.enrolmentVpdId).flatMap(block)
+      .recover {
+        case _ => Redirect(controllers.routes.JourneyRecoveryController.onPageLoad())
+      }
```
ian-hmrc: Should we be going to the journey recovery controller here? That should probably only be done for specific situations that we know the user can recover from. I'd be more inclined to just let the failure propagate out and be handled by the Error Handler.

app/controllers/returns/submit/adjustments/SelectAdjustmentPeriodController.scala
```diff
@@ -0,0 +1,59 @@
...
+                                                  requireData: ReturnsDataRequiredAction,
+                                                  obligationService: ObligationService,
+                                                  val controllerComponents: MessagesControllerComponents,
+                                                  view: SelectAdjustmentPeriodView
+                                                )(using ExecutionContext) extends FrontendBaseController with I18nSupport {
+
+  def onPageLoad(year: Option[Int]): Action[AnyContent] =
+    (identify andThen returnsEnabledAction andThen getData andThen requireData).async { implicit request =>
+
+      obligationService.getObligations(request.enrolmentVpdId).map { obligationsResponse =>
```
ian-hmrc: Do you want to consider using `obligationsService.getObligationsDirectly()` to save passing around the response and the having to do the same filtering down in `PeriodSelectHelper.filterFulfilledWithinThreeYears()`?

The same would apply to the usage in `AdjustmentVolumeWithTypeController` and `AdjustmentCheckYourAnswersController`.

I added `obligationsService.getObligationsDirectly()` recently to wean us off the other mechanism. If it's use gets spread enough, we can remove the old exposing of the `ObligationsResponse` and simplify a bunch of code.

app/navigation/ReturnsNavigator.scala
```diff
@@ -44,6 +45,9 @@ class ReturnsNavigator @Inject()(
     case DeclareSpoiltProductsPage      => ua  => declareSpoiltProductsPageRoutes(ua, periodKey)
     case AddSpoiltAdjustmentPage        => ua  => addSpoiltAdjustmentPageRoutes(ua, periodKey)
     case SpoiltVolumeByPeriodPage       => _   => withPeriod(controllers.returns.submit.spoilt.routes.AddSpoiltAdjustmentController.onPageLoad(NormalMode), periodKey)
+    case DeclareAdjustmentPage  => ua  => declareAdjustmentQuestionPageRoutes(ua, periodKey)
```
ian-hmrc: Tiny one, but the formatting doesn't line up here.

app/navigation/ReturnsNavigator.scala
```diff
@@ -44,6 +45,9 @@ class ReturnsNavigator @Inject()(
...
+    case AdjustmentListPage             => _   => withPeriod(controllers.returns.submit.adjustments.routes.AdjustmentCheckYourAnswersController.onPageLoad(), periodKey)
+    case AddAnotherAdjustmentPage       => ua  => addAnotherAdjustmentPageRoutes(ua, periodKey)
     case DeclarationPage                => _   => withPeriod(controllers.returns.submit.routes.ConfirmationController.onPageLoad(), periodKey)
     case _                              => _   => Call(GET, BtaLink(config))
```
ian-hmrc, line 52: Another tiny one, but would a line of whitespace between each of the tasks just help readability here? It's a bit of a solid block of code otherwise.

Maybe we could re-order them too, to follow the order of the tasks on the page? Again, might just help reading.

conf/app.routes
```diff
@@ -84,6 +84,21 @@ POST       /complete-return/adjustment/add-another-spoilt-adjustment
 GET        /complete-return/adjustment/change-add-another-spoilt-adjustment                 controllers.returns.submit.spoilt.AddSpoiltAdjustmentController.onPageLoad(mode: Mode = CheckMode)
 POST       /complete-return/adjustment/change-add-another-spoilt-adjustment                 controllers.returns.submit.spoilt.AddSpoiltAdjustmentController.onSubmit(mode: Mode = CheckMode)
 
+GET        /complete-return/adjustment/declare-adjustments                                  controllers.returns.submit.adjustments.DeclareAdjustmentQuestionController.onPageLoad(mode: Mode = NormalMode)
+POST       /complete-return/adjustment/declare-adjustments                                  controllers.returns.submit.adjustments.DeclareAdjustmentQuestionController.onSubmit(mode: Mode = NormalMode)
+GET        /complete-return/adjustment/change-declare-adjustments                           controllers.returns.submit.adjustments.DeclareAdjustmentQuestionController.onPageLoad(mode: Mode = CheckMode)
+POST       /complete-return/adjustment/change-declare-adjustments                           controllers.returns.submit.adjustments.DeclareAdjustmentQuestionController.onSubmit(mode: Mode = CheckMode)
```
ian-hmrc, line 90: I think we could de-duplicate these routes for the `CheckMode` by passing the query parameter directly rather than defaulting it against another URL. The `/view-your-returns` URL above is an example of how that can work.

app/utils/ReturnsDateUtils.scala
```diff
@@ -16,27 +16,29 @@
...
+import models.identifiers.PeriodKey
+import models.obligations.ObligationsResponse
 import play.api.i18n.Messages
 
 import java.time.{LocalDate, Month, Year}
 
 object ReturnsDateUtils {
-
+  // scalafix:off DisableSyntax.throw
   val month: Month = LocalDate.now().getMonth
```
ian-hmrc: Can we inject a clock here so that we can control time in tests?

conf/app.routes
ghousedev: I feel this would not be a good change to make, /view-your-returns is a read only route with no Mode needed so I am not sure that example is comparable.

This way of enabling the check mode is the scaffold standard way of doing things and is present in every repository and even in every controller that requires a Mode in this repository, it is an established pattern and could make maintenance more difficult in the future.

Using a query parameter here also loses compile time type safety.

app/utils/ReturnsDateUtils.scala
ghousedev: Yes, that makes sense.

app/controllers/returns/submit/adjustments/AdjustmentCheckYourAnswersController.scala
ghousedev: Have made some changes here, let me know what you think.

conf/app.routes
ian-hmrc: Yes, after adding the comment I made a small spike around this. It's not as straight-forward as I was hoping because the support for query parameters in Play doesn't seem to be very strong, particularly around the reverse routing. We can leave this as it is for now.

app/navigation/ReturnsNavigator.scala
ghousedev: I have changed the order to reflect the task page.

app/controllers/returns/submit/adjustments/AdjustmentCheckYourAnswersController.scala
ian-hmrc: That looks good. :)

app/controllers/returns/submit/adjustments/AdjustmentCheckYourAnswersController.scala
ghousedev: Have introduced a service to handle this.

app/controllers/returns/submit/adjustments/SelectAdjustmentPeriodController.scala
ian-hmrc: Cool. I can see that's been done.

ian-hmrc (approved):
LGTM :)

PR #125, opened 2026-07-09
Vpd 1862 - Build 'Over/under adjustments' reason page

craigemmerson:


test/navigation/ReturnsNavigatorSpec.scala
```diff
@@ -160,6 +161,29 @@ class ReturnsNavigatorSpec extends SpecBase {
         navigator.nextPage(AddAnotherAdjustmentPage, NormalMode, ua).url mustBe s"${controllers.returns.submit.routes.TaskListController.onPageLoad().url}?period=$periodKey"
       }
 
+      "must go from AddAnotherAdjustmentPage to AdjustmentReason when no more adjustments to make and an adjustment volume is 1000ml or more" in {
+        val ua = returnsUserAnswers
```
ian-hmrc: This needs to be done if it is over £1000, not 1000ml.

test/forms/returns/AdjustmentReasonFormProviderSpec.scala
```diff
@@ -48,5 +48,26 @@ class AdjustmentReasonFormProviderSpec extends StringFieldBehaviours {
       fieldName,
       requiredError = FormError(fieldName, requiredKey)
     )
+
+    "bind a string that is below the maximum length" in {
+      val validData = "a" * 50
+      val result = form.bind(Map(fieldName -> validData)).apply(fieldName)
+      result.value.value mustBe validData
+      result.errors mustBe empty
+    }
```
ian-hmrc, line 57: This test would not have shown the problem of the spaces being removed. Rather than just having 50 'a's in a row, could we have a realistic phrase that goes end-to-end? Maybe something of the form: 'April 2027: Error in warehouse data recording'. And show that it comes through correctly?

ian-hmrc (approved):
LGTM :)

PR #128, opened 2026-07-11
VPD-1775 duty rate per10ml

ian-hmrc:
Change the duty rate configuration to be in pence-per-10ml instead of pence-per-ml. This allows duty rates to be specified to the full resolution described in the Return Data Items v2.11.


ghousedev (approved):
👍

PR #129, opened 2026-07-11
VPD-1775 duty rate domain type

ian-hmrc:
Introduce a domain type for the Duty Rate to address the primitive obsession and duplication code smells.

This centralises the duty calculations in one place, giving them consistency.

To do this, I've renamed the existing `DutyRate` case class to be called `ConfigDutyRate`, which otherwise allows the config parsing code to be left alone. In particular, I've made `DutyRateService` return the new `DutyRate` type instead of primitives, and then pushed that further through the app.

It also fixes a bunch of tests that were specifying the duty rate incorrectly as it was just a primitive (previously there were different rates in different places, in pounds or pence and per ml or per litre, and it was easy to confuse them). Now there is a domain type, this cannot happen.


ghousedev (approved):
👍

app/viewmodels/returns/submit/adjustments/AdjustmentCheckYourAnswersViewModel.scala
```diff
@@ -84,11 +80,11 @@ object AdjustmentCheckYourAnswersViewModel {
   private def totalDutyForType(
                                 adjustments: Seq[AdjustmentEntry],
                                 adjustmentType: AdjustmentType,
-                                dutyRates: Map[String, BigDecimal]
+                                dutyRates: Map[PeriodKey, DutyRate]
                               ): BigDecimal =
     adjustments
       .filter(_.adjustmentType == adjustmentType)
-      .map(adjustment => calculateDuty(adjustment.volumeInMl, dutyRates.getOrElse(adjustment.period.toString, BigDecimal(0))))
+      .map(adjustment => dutyRates.get(adjustment.period).map(_.calculateDuty(adjustment.volumeInMl)).getOrElse(BigDecimal(0)))
```
ghousedev, line 87: Is it worth adding a BigDecimal("0") to the ReturnsConstants file.

ian-hmrc: TBH, I'd rather these defaults go away rather than giving them a nice home. This is what we should get with the validation if we introduce it. Can we wait and see how that goes before doing this?

PR #130, opened 2026-07-13
Vpd 1888 - Mini CYA spoilt page

craigemmerson:


ghousedev (dismissed):
👍

app/controllers/returns/submit/spoilt/SpoiltCheckYourAnswersController.scala
```diff
@@ -0,0 +1,106 @@
...
+import play.api.mvc.{Action, AnyContent, MessagesControllerComponents, Result}
+import services.returns.{ReturnsUserAnswersService, SpoiltCheckYourAnswersService}
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendBaseController
+import views.html.returns.submit.spoilt.SpoiltCheckYourAnswersView
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class SpoiltCheckYourAnswersController @Inject()(
```
ian-hmrc, line 37: If we're renaming from `AddSpoiltAdjustmentController` to `SpoiltCheckYourAnswersController`, then should we also rename the `AddSpoiltAdjustmentPage` object?

app/controllers/returns/submit/spoilt/SpoiltCheckYourAnswersController.scala
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  def onSubmit(): Action[AnyContent] = (identify andThen returnsEnabledAction andThen getData andThen requireData).async {
+    implicit request =>
+
+      val declareSpoiltProducts = request.userAnswers.get(DeclareSpoiltProductsPage)
+      val spoiltList = request.userAnswers.get(SpoiltVolumeByPeriodPage)
+
+      spoiltCheckYourAnswersService
+        .buildViewModel(declareSpoiltProducts, spoiltList, request.periodKey, request.enrolmentVpdId)
```
ian-hmrc: The view model is being built in all cases here, when it is only used for failures. Might this be moved more locally to where it is needed?

craigemmerson: this was some of the things I was talking about around the adjustments doing the same thing across controllers , I can change how this behaves inline with your comment but then the other CYA controller does not aline so worth noting

app/controllers/returns/submit/spoilt/SpoiltCheckYourAnswersController.scala
ian-hmrc: Cool :)

app/controllers/returns/submit/spoilt/SpoiltCheckYourAnswersController.scala
ian-hmrc: Cool. I can see that's been addressed. 👍

ian-hmrc (approved):
LGTM :)

PR #132, opened 2026-07-16
Vpd 1816 - build 'remove adjustments screen'

craigemmerson:


app/controllers/returns/submit/adjustments/RemoveAdjustmentController.scala
```diff
@@ -0,0 +1,108 @@
...
+            case None => Redirect(controllers.routes.JourneyRecoveryController.onPageLoad())
+          }
+      }
+  }
+
+  def onSubmit(): Action[AnyContent] = (identify andThen returnsEnabledAction andThen getData andThen requireData).async {
+    implicit request =>
+      withPeriodKey(ReturnsConstants.QUERY_PARAM_ADJUSTMENT_PERIOD) { adjustmentPeriod =>
+        adjustmentCheckYourAnswersService
+          .buildRemoveViewModel(request.userAnswers.get(AdjustmentListPage), adjustmentPeriod, request.enrolmentVpdId)
```
ian-hmrc: We're building the view model here which is only used in the case of form errors. It involves making a request for the obligations, which will be unnecessary if the form has been filled out correctly. Moreover, should getting the obligations fail, the whole action for this submission will be abandoned when it could have succeeded for the happy path where a radio button has been selected. Can we flip the order of these around to check the form errors first, and only build the view model if there are failures? This should also be done for the Spoilt removal too.

ian-hmrc: I've just played with this, and once done, you can also replace the `flatMap()` of the response of building the VM with a `map()` and unwind some wrapping in `Future.successful()`.

app/controllers/returns/submit/adjustments/RemoveAdjustmentController.scala
```diff
@@ -0,0 +1,108 @@
...
+                confirmed =>
+                  val resultAnswers =
+                    if (confirmed) {
+                      for {
+                        updatedAnswers <- Future.fromTry(removeEntry(request.userAnswers, adjustmentPeriod))
+                        _              <- sessionRepository.set(updatedAnswers)
+                      } yield updatedAnswers
+                    } else {
+                      Future.successful(request.userAnswers)
+                    }
```
ian-hmrc: This work to build `resultAnswers` makes this controller method bigger than it ought to be, and also the name `resultAnswers` isn't speaking to me at the level of what's going on here. Can we extract a function for this (maybe `handleRemovalRequest()` and then I don't think we need the `val`.

We could go further and move the extracted method out of the controller and into a service. This would align better with the Ports and Adapters architectural style we talked about where the Controller just acts as a thin adaptor layer, translating the incoming request from being in HTTP to being in domain terms. This moves business logic out of the controller where it's harder to test, and into a place where it's easier to unit-test.

app/controllers/returns/submit/spoilt/RemoveSpoiltAdjustmentController.scala
```diff
@@ -52,35 +53,46 @@ class RemoveSpoiltAdjustmentController @Inject()(
...
 
   def onSubmit(): Action[AnyContent] = (identify andThen returnsEnabledAction andThen getData andThen requireData).async {
     implicit request =>
       withPeriodKey(ReturnsConstants.QUERY_PARAM_SPOILT_PERIOD) { spoiltPeriod =>
-        form.bindFromRequest().fold(
-          formWithErrors =>
-            Future.successful(BadRequest(view(request.periodKey, spoiltPeriod, formWithErrors))),
+        spoiltCheckYourAnswersService
+          .buildRemoveViewModel(request.userAnswers.get(SpoiltVolumeByPeriodPage), spoiltPeriod, request.enrolmentVpdId)
+          .flatMap {
```
ian-hmrc: This is the same as the under/over adjustments. We should only build the view model if the validation has failed.

test/controllers/returns/submit/adjustments/RemoveAdjustmentControllerSpec.scala
```diff
@@ -0,0 +1,266 @@
...
+
+      val application = applicationBuilder(returnsUserAnswers = Some(userAnswers))
+        .overrides(
+          bind[ObligationService].toInstance(stubbedObligationService),
+          bind[DutyRateService].toInstance(stubbedDutyRateService)
+        )
+        .build()
+
+      running(application) {
+        val request = FakeRequest(GET, removeAdjustmentRoute)
```
ian-hmrc: I find it harder to read tests if they don't make the parameters easily readable. This removal is for October 2027, and I'd expect to see that here. Making that happen is not simple because of the way we build URLs from reverse controllers, and the framework injects a mutable prefix. I've put together a patch and sent it on the slack channel regarding this. Happy to chat this through at some point if you want to discuss the pros and cons.

test/services/returns/SpoiltCheckYourAnswersServiceSpec.scala
ian-hmrc: `DUTY_RATE` is very abstract here and doesn't make it easy to see why the duty calculation is working out below. If it were renamed to `TEN_POUNDS_PER_10ML`, it would make the test more readable locally.

We could even raise the game further by introducing a test helper function like `poundsPer10ml("10.00")`, or even better, especially if we eventually add a domain type for monetary values, something like `per10Ml(GBP("10.00"))`.

test/services/returns/AdjustmentCheckYourAnswersServiceSpec.scala
```diff
@@ -177,4 +177,64 @@ class AdjustmentCheckYourAnswersServiceSpec extends SpecBase with MockitoSugar w
...
+        adjustmentType = AdjustmentType.OverDeclared,
+        volumeInMl = BigDecimal(1000)
+      )
+      val adjustmentList = AdjustmentList(Seq(adjustmentEntry))
+      val obligationForAdjustment = openObligation(adjustmentPeriodKey)
+
+      when(mockObligationService.getObligationsDirectly(eqTo(vpdId))(using any()))
+        .thenReturn(Future.successful(Seq(obligationForAdjustment)))
+      when(mockDutyRateService.getDutyRateForDate(any()))
+        .thenReturn(DUTY_RATE_PENCE_PER_10ML)
```
ian-hmrc: This is similar to the situation in SpoiltCheckYourAnswersServiceSpec

test/services/returns/AdjustmentCheckYourAnswersServiceSpec.scala
```diff
@@ -177,4 +177,64 @@ class AdjustmentCheckYourAnswersServiceSpec extends SpecBase with MockitoSugar w
...
+        .thenReturn("December 2026")
+
+      val result = service.buildRemoveViewModel(
+        adjustmentList = Some(adjustmentList),
+        adjustmentPeriod = adjustmentPeriodKey,
+        vpdId = vpdId
+      ).futureValue
+
+      result.value.rows.size mustBe 4
+    }
```
ian-hmrc, line 156: This test, and the other tests in this file are quite long and contain a lot of boilerplate. I have an example refactoring of this test where I can take it from 23 lines to 14 lines by simply extracting test helpers. I've also done this for the rest of the file.

Moreover, by pulling way duplication and boilerplate, I'm able to concentrate more clearly on domain concepts. With this, and the inlining of the overly abstract test data, I think the result is far more readable. Here's a sample:

```
    "must return Some with the details rows when the entry and its obligation are found" in {
      stubObligations(Seq(openObligation(february2024)))
      stubDutyRate(TEN_POUNDS_PER_10ML)

      val result = service.buildRemoveViewModel(
        adjustmentList = Some(AdjustmentList(Seq(
          overDeclaredAdjustment(february2024, ml(1000))
        ))),
        adjustmentPeriod = february2024,
        vpdId = vpdId
      ).futureValue

      result.value.rows.size mustBe 4
    }
```

I can take you through this at some point if you like.

craigemmerson: have updated the spec to use the helpers we discussed

ian-hmrc: Cool. That looks much more readable to me.

ian-hmrc (approved):

PR #134, opened 2026-07-17
VPD-1887: Add over/under adjustments to view individual return page

ghousedev:


app/viewmodels/returns/view/AdjustmentsSectionBuilder.scala
```diff
@@ -0,0 +1,196 @@
...
+import models.returns.view.{OverDeclaration, UnderDeclaration}
+import play.api.i18n.Messages
+import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Text
+import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.*
+import utils.{CssConstants, CurrencyFormatter, ReturnsDateUtils}
+
+final case class AdjustmentsSectionBuilder(
+  overDeclaration: Option[OverDeclaration],
+  underDeclaration: Option[UnderDeclaration],
+  obligations: ObligationsResponse,
```
ian-hmrc: Could we use `ObligationsService.getObligationsDirectly()` to save having to pass the whole response around? We only need the `Seq[ObligtionDetails]`, right?

app/viewmodels/returns/view/ViewIndividualReturnViewModel.scala
```diff
@@ -38,16 +37,17 @@ case class ViewIndividualReturnViewModel(
                                           nilReturn: Boolean,
```
ian-hmrc: The `nilReturn` parameter is unused here and should be removed.

ghousedev: It is used in the view
`    @if(!vm.nilReturn) {
        @h2(messages("viewIndividualReturn.h2.dutyDetails"))
        @p(Seq(Text(messages("viewIndividualReturn.dutyDetails", vm.dutyRate))))
    } else {
        @p(Seq(Text(messages("viewIndividualReturn.nil.dutyDetails"))))
    }`

ian-hmrc: Looking further at how `nilReturn` is setup in `ViewIndividualReturnViewModel`, it is incorrect about the definition of a nil return, assuming a return is nil if the duty section is zero.

A nil return is when the _total charge_ from the return is zero. This can be because all elements are zero, or because charges are equally matched by reimbursements from Spoilt and Over declared. In this case no charge reference is produced as there is no charge (or reimbursement) to make.

Other than it's unnecessary use here, `nilReturn` is used in the view to:
* To show/hide the charge reference if it is present. I think this could be changed to make the charge reference and Option, and map it in the view. That will make it clearer that it can be missing.
* To show/hide the _'The amount is based on the total duty due on:'_, _'the vaping products you declared'_ and _'all applicable adjustments you declared'_ content at the top of the page. Looking at this more closely, I'm not sure this behaviour is correct. I think this should not be dynamic at all. We've had a discussion. with the UxD and decided to leave it all showing all the time.
* To show/hide the 'Final duty summary', and the paragraph that includes the duty rate. Again, this has been discussed with UxD and we should keep the h2 fixed (not dynamic) for this, and only switch out the paragraph if there is no duty declared (i.e. if we don't have a duty rate show).

ghousedev: Have made some changes to improve how this content is rendered

app/viewmodels/returns/view/ViewIndividualReturnViewModel.scala
```diff
@@ -16,53 +16,53 @@
...
     val success = returnsData.success
 
-    val dutyRate = returnsData.success.vapingProductsProduced
+    val dutyRateValue = returnsData.success.vapingProductsProduced
       .flatMap(_.returns.headOption)
       .map(_.dutyRate)
 
-    val isNilReturn = dutyRate.forall(_ == zeroValue)
+    val dutyRateFormatted = dutyRateValue
+      .filter(_ != zeroValue)
```
ian-hmrc: I don't think we need to filter for the duty rate not being zero. Under no circumstances should the duty rate be zero, and this makes me uneasy that somehow we might be expecting it to be zero.

ian-hmrc (approved):
LGTM :)

PR #135, opened 2026-07-20
[VPD-1736] update bounced email value

craigemmerson:


test/services/contactPreference/EmailVerificationServiceSpec.scala
```diff
@@ -202,6 +203,27 @@ class EmailVerificationServiceSpec extends SpecBase {
       }
     }
 
+    "must submit a PaperlessPreferenceSubmission with bouncedEmail set to Some(false) when successful" in new Setup {
```
ian-hmrc: Small one, but this test name is quite technical. I think it could be more domain-centric. For example _submitVerifiedEmail_ 'must always clear the bounced email flag (even if not set) so the manufacturer can receive emails again'.

craigemmerson: im happy with that name. I can push that now so this can go in with the backend pr

ian-hmrc (approved):
LGTM :)

PR #136, opened 2026-07-21
VPD-1843: Consistent captions

ghousedev:


app/controllers/returns/submit/DeclarationController.scala
```diff
@@ -52,31 +55,44 @@ class DeclarationController @Inject()(
...
+      obligationService.getObligationsDirectly(request.enrolmentVpdId).map { obligations =>
+        val preparedForm = request.userAnswers.get(DeclarationPage) match {
+          case None => form
+          case Some(value) => form.fill(value)
+        }
+        
+        val periodDisplay = returnsDateUtils.formatPeriodDisplay(request.periodKey, obligations)
+        val parts = periodDisplay.split(" ")
+        val periodMonth = parts.head
+        val periodYear = parts.last
```
ian-hmrc: This feels like a bit of a hack to be splitting and selecting the parts? Did you consider spitting the formatting into something that could generate a month and year tuple or case class? That would feel more solid than manipulating the text.

test/controllers/returns/submit/DeclarationControllerSpec.scala
```diff
@@ -44,41 +45,68 @@ class DeclarationControllerSpec extends SpecBase with MockitoSugar {
...
         running(application) {
           val request = FakeRequest(GET, controllers.returns.submit.routes.DeclarationController.onPageLoad().url)
 
           val result = route(application, request).value
 
           val view = application.injector.instanceOf[DeclarationView]
+          val periodDisplay = returnsDateUtils.formatPeriodDisplay(periodKey, obligationDataSingleOpen)
+          val parts = periodDisplay.split(" ")
+          val periodMonth = parts.head
+          val periodYear = parts.last
```
ian-hmrc: This is a duplicate of the app logic in the tests, which I'd rather we avoided. Making the values more apparent in the test can allow us to avoid this issue. In this example I've surfaced the `periodKey` as `june2026` (in the user answers), and then I can simply assert against the strings we expect for that period.

```
      "must return OK and the correct view for a GET" in {

        val mockObligationService = mock[ObligationService]
        when(mockObligationService.getObligationsDirectly(any())(using any()))
          .thenReturn(Future.successful(obligationDetailsSingleOpen))

        val application = applicationBuilder(returnsUserAnswers = Some(returnsUserAnswers.copy(periodKey = june2026.value)))
          .overrides(bind[ObligationService].toInstance(mockObligationService))
          .build()

        running(application) {
          val request = FakeRequest(GET, controllers.returns.submit.routes.DeclarationController.onPageLoad().url)

          val result = route(application, request).value

          val view = application.injector.instanceOf[DeclarationView]

          status(result) mustEqual OK
          contentAsString(result) mustEqual view(june2026, form, "June", "2026")(request, messages(application)).toString
        }
      }
```

app/controllers/returns/submit/DeclarationController.scala
ghousedev: Have added a case class for this.

test/controllers/returns/submit/DeclarationControllerSpec.scala
ghousedev: The tests have changed from the introduction of the case class, see if you still feel the same.

ian-hmrc: TBH, I still feel the same. :)

ian-hmrc: My example above put `june2026` into the userAnswers. That make it present and easily visible in the tests (though maybe I should have done more about the long line in the example above - mea culpa). I think this is key to the readability of the tests. You can see June 2026 going in, and you can see it formatted correctly coming out. Sorry to be a pain, but could we add that in too please?

ghousedev: Sure, should be all done now.

ian-hmrc (approved):
LGTM :)

PR #138, opened 2026-07-24
Vpd 1920 - View payments empty states

craigemmerson:


app/viewmodels/payments/ViewPaymentsViewModel.scala
```diff
@@ -28,27 +28,32 @@ import java.time.format.DateTimeFormatter
...
 
-  private val DATE_FORMATTER = DateTimeFormatter.ofPattern("MMMM yyyy")
+  private val DATE_FORMATTER = DateTimeFormatter.ofPattern("d MMMM yyyy")
 
   private val govukTag = GovukTag()
 
-  def apply(paymentOption: Seq[OutstandingPayment])(implicit messages: Messages): ViewPaymentsViewModel = {
-    val totalOwed = CurrencyFormatter.currencyFormat(paymentOption.map(_.amountDue).sum)
+  def apply(payments: PaymentsResponse)(implicit messages: Messages): ViewPaymentsViewModel = {
+    val totalOwed = CurrencyFormatter.currencyFormat(payments.outstanding.map(_.amountDue).sum)
```
ian-hmrc: We should not be calculating the total due. This should come from the API. VPD-1956 states in the data mapping that it should come from `financialData.totalisation.regimeTotalisation.totalAccountBalance`. ETMP should give us this value and we should just display it. If we calculate it, there's a risk that we display something incorrect to the user.

app/controllers/payments/ViewPaymentsController.scala
```diff
@@ -39,13 +39,13 @@ class ViewPaymentsController @Inject()(
 )(using ExecutionContext) extends FrontendBaseController with I18nSupport with Logging {
 
   def onPageLoad: Action[AnyContent] = identify.async { implicit request =>
```
ian-hmrc: We also only allow access to this endpoint if the Phase 2 feature switch is enabled.

app/viewmodels/payments/ViewPaymentsViewModel.scala
ghousedev: Have passed totalAccountBalance through

ian-hmrc (approved):
LGTM :)

PR #139, opened 2026-07-24
VPD-1886: Add spoilt to CYA, Update CYA design

ghousedev:


app/controllers/returns/submit/spoilt/SpoiltCheckYourAnswersController.scala
```diff
@@ -50,62 +50,66 @@ class SpoiltCheckYourAnswersController @Inject()(
...
-  private def redirectToNextPageWithoutAddingAnother(request: ReturnsDataRequest[AnyContent])
+  private def redirectToNextPageWithoutAddingAnother(request: ReturnsDataRequest[AnyContent], mode: Mode)
                                                     (using HeaderCarrier): Future[Result] = {
-    for {
-      updatedAnswers <- Future.fromTry(request.userAnswers.set(SpoiltCheckYourAnswersPage, false))
-      _              <- sessionRepository.set(updatedAnswers)
-    } yield Redirect(navigator.nextPage(SpoiltCheckYourAnswersPage, NormalMode, updatedAnswers))
+    mode match {
+      case CheckMode =>
+        Future.successful(Redirect(controllers.returns.submit.routes.CheckYourAnswersController.onPageLoad().url + s"?period=${request.periodKey.value}"))
```
ian-hmrc: Is there a reason why the URL to redirect to is inlined here rather than going via the navigator?

The user answers are not getting updated here as they are in the case of normal mode. Isn't this an omission?

app/controllers/returns/submit/DeclareDutyCheckAnswersController.scala
```diff
@@ -39,18 +40,21 @@ class DeclareDutyCheckAnswersController @Inject()(
...
     }
   }
 
-  def onSubmit: Action[AnyContent] = (identify andThen returnsEnabled andThen getData andThen requireData) { implicit request =>
-    Redirect(controllers.returns.submit.routes.TaskListController.onPageLoad().url + s"?period=${request.periodKey.value}")
+  def onSubmit(mode: Mode = NormalMode): Action[AnyContent] = (identify andThen returnsEnabled andThen getData andThen requireData) { implicit request =>
+    mode match {
+      case CheckMode => Redirect(controllers.returns.submit.routes.CheckYourAnswersController.onPageLoad().url + s"?period=${request.periodKey.value}")
+      case NormalMode => Redirect(controllers.returns.submit.routes.TaskListController.onPageLoad().url + s"?period=${request.periodKey.value}")
+    }
```
ian-hmrc, line 58: Would this routing decision not be better if it were in the Navigator?

app/viewmodels/returns/submit/spoilt/SpoiltCheckYourAnswersViewModel.scala
```diff
@@ -146,16 +155,18 @@ object SpoiltCheckYourAnswersViewModel {
...
     SummaryListRow(
       key = Key(content = Text(messages("returns.declareSpoiltProducts.question"))),
       value = Value(content = Text(messages(if (declared) "site.yes" else "site.no"))),
       actions = Some(Actions(items = Seq(
         ActionItem(
-          href = buildSpoiltUrl(
-            controllers.returns.submit.spoilt.routes.DeclareSpoiltProductsController.onPageLoad(NormalMode).url,
-            currentPeriodKey
-          ),
+          href = s"${controllers.returns.submit.spoilt.routes.DeclareSpoiltProductsController.onPageLoad(changeMode).url}?period=${currentPeriodKey.value}",
```
ian-hmrc: There are a couple of points in this file where `buildSpoiltUrl()` has been replaced with inline construction of the URL. These seem to be where the `Mode` is not required. Would it make more sense for the function to take an `Option[Mode]` parameter instead, so that all the adding of query parameters can be done in one place? Inlining like this duplicates the parts that are inlined.

One further though is that the function is just adding query parameters rather than building the URL. Maybe it could be named better to reflect this?

app/viewmodels/returns/submit/spoilt/SpoiltCheckYourAnswersViewModel.scala
```diff
@@ -102,24 +109,26 @@ object SpoiltCheckYourAnswersViewModel {
...
 
     val cardActions = Seq(
       ActionItem(
         href = buildSpoiltUrl(
-          controllers.returns.submit.spoilt.routes.RemoveSpoiltAdjustmentController.onPageLoad().url,
+          controllers.returns.submit.spoilt.routes.RemoveSpoiltAdjustmentController.onPageLoad(mode).url,
           currentPeriodKey,
-          Some(entry.periodKey)
+          Some(entry.periodKey),
+          mode
```
ian-hmrc: The `mode` is being added as a query parameter here, and also being passed forward through the reverse routes. Do we need both? The `RemoveSpoiltAdjustmentController` does pick up the query parameter.

If this is the case, then the mode could be removed from `buildUrl()` and the comments below about `buildUrl()` being inlined could be reverted.

app/controllers/returns/submit/spoilt/SpoiltCheckYourAnswersController.scala
ghousedev: Fixed that

app/controllers/returns/submit/DeclareDutyCheckAnswersController.scala
ghousedev: We have followed the pattern of not using the Navigator class for navigation that is not driven by a Page object and related value held in Mongo.  This logic could still be moved to a function in the Navigator if we feel that is a better approch.

ian-hmrc: I'm fine with following the pattern we have. We might want to keep an eye on this in the future, but for now we can keep this as is.

app/viewmodels/returns/submit/spoilt/SpoiltCheckYourAnswersViewModel.scala
ian-hmrc: Cool. I can see the query param has been removed.

app/viewmodels/returns/submit/CheckYourAnswersViewModel.scala
```diff
@@ -31,8 +31,8 @@ import views.html.components.Paragraph
 
 case class CheckYourAnswersViewModel(
                                       declareDutyCard: (String, SummaryList, Option[Seq[ActionItem]]),
-                                      spoiltProductsCard: (String, SummaryList, Option[Seq[ActionItem]]),
-                                      adjustmentsCard: (String, SummaryList, Option[Seq[ActionItem]]),
+                                      spoiltProductsCard: Option[(String, SummaryList, Option[Seq[ActionItem]])],
+                                      adjustmentsCard: Option[(String, SummaryList, Option[Seq[ActionItem]])],
                                       dutySuspendedCard: (String, SummaryList, Option[Seq[ActionItem]]),
```
ian-hmrc: I can see we're using the same tuple here for each card. Would a case class not help here?

```
case class CheckYourAnswersViewModel(
                                      declareDutyCard    :        (String, SummaryList, Option[Seq[ActionItem]]),
                                      spoiltProductsCard : Option[(String, SummaryList, Option[Seq[ActionItem]])],
                                      adjustmentsCard    : Option[(String, SummaryList, Option[Seq[ActionItem]])],
                                      dutySuspendedCard  :        (String, SummaryList, Option[Seq[ActionItem]]),
```

Also, the `Option[Seq[ActionItem]]` is only ever used with `Some`. I can see how it has been used in the view to omit the `<ul>` when there are no items, but is this strictly necessary?

ghousedev: The Actions are not shown or needed in other views that use the SummaryCard component, hence why it is an Option.

ian-hmrc (approved):
LGTM :)

PR #140, opened 2026-07-28
VPD-1957: View payments - Unallocated

ghousedev:


app/viewmodels/payments/ViewPaymentsViewModel.scala
```diff
@@ -42,15 +43,22 @@ object ViewPaymentsViewModel {
 
   def apply(payments: PaymentsResponse, returnsDateUtils: ReturnsDateUtils)(implicit messages: Messages): ViewPaymentsViewModel = {
     val totalOwed = payments.totalAccountBalance.getOrElse(BigDecimal(0))
+    val clearedYear = extractYearFromClearedPayments(payments.cleared, returnsDateUtils)
```
ian-hmrc: The current year shouldn't need to be extracted from the cleared payments like this. The year should be known from the request (for the first page, it's the current year). There will be a ticket coming to add the pagination, where for those pages their year should also be known from the request.

The ticket did say in the mappings table that this could be static content for now. Either we can revert to that, or possibly take the current year at a pinch?

ghousedev: I knew this would be a temporary way of adding the year, happy to change and have done so.  Once we add pagination we can pass the year through the apply method with minimal changes.

ian-hmrc (dismissed):
LGTM :)

ian-hmrc (approved):
LGTM :)

craigemmerson (approved):

PR #145, opened 2026-08-17
VPD-1864: Add support for insolvent status

ghousedev:


app/controllers/actions/CheckInsolvencyAction.scala
```diff
@@ -0,0 +1,57 @@
...
+    implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequestAndSession(request, request.session)
+
+    subscriptionConnector.getSubscriptionContactPreferences(request.enrolmentVpdId).map {
+      case Right(subscription) =>
+        subscription.insolvencyStatus match {
+          case Some(INSOLVENCY_STATUS_INSOLVENT) =>
+            logger.info(s"User with VpdId ${request.enrolmentVpdId.value} is insolvent, redirecting")
+            // TODO: Replace with actual insolvency page route when created
+            Left(Redirect(controllers.routes.JourneyRecoveryController.onPageLoad()))
+          case _ =>
```
ian-hmrc: The case handles the manufacturer being either Solvent, or the insolvency flag absent. It would be safer to only allow them access to the service if they are clearly Solvent. I'm sure the flag is always present, but I feel we should flip this logic just to be sure.

app/controllers/actions/CheckInsolvencyAction.scala
```diff
@@ -0,0 +1,57 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+trait CheckInsolvencyAction extends ActionRefiner[IdentifierRequest, IdentifierRequest]
+
+class CheckInsolvencyActionImpl @Inject()(
+                                           subscriptionConnector: SubscriptionConnector
+                                         )(implicit val executionContext: ExecutionContext)
+  extends CheckInsolvencyAction with Logging {
+
+  private val INSOLVENCY_STATUS_INSOLVENT = "Y"
```
ian-hmrc: Could an enum or sealed class be used to hold these 2 states? It would be clearer than the strings.

app/models/contactPreference/SubscriptionSummary.scala
```diff
@@ -24,7 +24,8 @@ case class SubscriptionSummary(
                                 emailVerification: Option[Boolean],
                                 bouncedEmail: Option[Boolean],
                                 correspondenceAddress: String,
-                                countryCode: Option[String]
+                                countryCode: Option[String],
+                                insolvencyStatus: Option[String] = None
```
ian-hmrc: This field is only used in `TestData`. It can be removed.

app/models/contactPreference/SubscriptionContactPreferences.scala
```diff
@@ -21,7 +21,8 @@ import models.emailverification.JsonHelpers._
 
 final case class SubscriptionContactPreferences(
                                                  paperlessPreference: Boolean,
-                                                 emailAddress: Option[String]
+                                                 emailAddress: Option[String],
+                                                 insolvencyStatus: Option[String]
```
ian-hmrc: This pollutes the model for contact preferences with data for insolvency. Could this be split out into its own model to avoid changing existing code?

test/controllers/actions/FakeCheckInsolvencyAction.scala
```diff
@@ -0,0 +1,32 @@
...
+
+package controllers.actions
+
+import data.TestData
+import models.requests.IdentifierRequest
+import play.api.mvc.*
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class FakeCheckInsolvencyAction extends CheckInsolvencyAction with TestData {
```
ian-hmrc: This does not need to include TestData

test/controllers/actions/FakeCheckInsolvencyActionInsolvent.scala
```diff
@@ -0,0 +1,33 @@
...
+package controllers.actions
+
+import data.TestData
+import models.requests.IdentifierRequest
+import play.api.mvc.*
+import play.api.mvc.Results.Redirect
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class FakeCheckInsolvencyActionInsolvent extends CheckInsolvencyAction with TestData {
```
ian-hmrc: This does not need to include TestData

app/models/contactPreference/SubscriptionContactPreferences.scala
ian-hmrc: Could this be spilt out from the Contact Prefs stuff? I can't see that being done yet.

ian-hmrc: Cool. Can see that's been done now.

ian-hmrc (approved):
LGTM :)

PR #146, opened 2026-08-18
Vpd 1982 - update confirmation page

craigemmerson:


app/viewmodels/returns/submit/ConfirmationViewModel.scala
```diff
@@ -108,37 +109,66 @@ object ConfirmationViewModel extends CurrencyFormatter {
...
+      classes = s"govuk-body ${CssConstants.paddingBottom2}"
+    )
 
-    val directDebitLink = p(Seq(
-      HtmlContent(
+    val firstBulletItem = if (isChapsRequired) {
+      p(Seq(HtmlContent(
+        link(
+          id = "chaps-payment-link",
+          href = "https://www.gov.uk/guidance/hmrc-bank-account-details",
```
craigemmerson: This link I not real yet, we will update when we get the right link

app/viewmodels/returns/submit/ConfirmationViewModel.scala
```diff
@@ -108,37 +109,66 @@ object ConfirmationViewModel extends CurrencyFormatter {
     val list = ListWithLinks()
     val link = Link()
 
+    val isChapsRequired = dutyDue >= CHAPS_PAYMENT_THRESHOLD
```
ian-hmrc: The reason for requiring payment by CHAPS is because £20M is above the Direct Debit threshold. Maybe the constant could be renamed here to be clearer about the reason for this? `val isChapsRequired = dutyDue >= MAX_DIRECT_DEBIT_THRESHOLD`

ghousedev (dismissed):
👍

ghousedev (approved):
👍

app/viewmodels/returns/submit/ConfirmationViewModel.scala
ian-hmrc: Cool. I can see that's been done.

app/viewmodels/returns/submit/ConfirmationViewModel.scala
ian-hmrc: Understood. We've got this captured on our phase 2 go-live list.

PR #147, opened 2026-08-20
VPD-1994: Update obligations API

ghousedev:


app/services/returns/ObligationService.scala
```diff
@@ -31,14 +31,14 @@ class ObligationService @Inject()(obligationsConnector: ObligationsConnector)
   def getObligations(vpdId: VpdId)(using HeaderCarrier): Future[Seq[ObligationDetails]] =
     obligationsConnector.getObligations(vpdId).map { response =>
       response.obligation
-        .map(_.obligationDetails)
+        .flatMap(_.obligationDetails)
```
ian-hmrc: Combining all `ObligationItems` in this manner into a list of just `ObligationDetails` would conflate obligations for different regimes that the manufacturer is subscribed to. They could be on ADR as well as VPD and pulling back obligations both of these. I realise we'll get away with this right now as we're specifically requesting with the X-Regime of VPD, but I think it would be more correct to filter the response for `ObligationItems` that have `referenceType` ZVPD.

TBH, I had it in mind that we would do that in `vaping-duty` and just present a simple list of obligations to the FE. That would seem to make good use of the BE. If that's a bit much to do right now, I'm happy to pick this up as a future piece of work if required.

ghousedev: Good point, have added filters for both the regime and VPDID.

app/services/returns/ObligationService.scala
```diff
@@ -17,31 +17,40 @@
...
-  def getObligations(vpdId: VpdId)(using HeaderCarrier): Future[ObligationsResponse] =
-    obligationsConnector.getObligations(vpdId)
-  
   def getObligationByPeriodKey(vpdId: VpdId, periodKey: PeriodKey)
                               (using HeaderCarrier): Future[Option[ObligationDetails]] =
     obligationsConnector.getObligations(vpdId).map { response =>
-      response.obligation
-        .map(_.obligationDetails)
+      filterObligationsByVpdId(response.obligation, vpdId)
         .find(_.periodKey == periodKey.toString)
```
ian-hmrc: Can we reduce the duplication here by re-using `getObligations()`?
```
getObligations(vpdId).map(_.find(_.periodKey == periodKey.toString))
```

test/controllers/returns/submit/adjustments/AdjustmentVolumeWithTypeControllerSpec.scala
```diff
@@ -69,9 +69,9 @@ class AdjustmentVolumeWithTypeControllerSpec extends SpecBase with MockitoSugar
       val mockObligationService = mock[ObligationService]
       val mockDutyRateService = mock[DutyRateService]
       val mockVolumePrecisionService = mock[VolumePrecisionService]
-      val obligationDetails = obligations(Seq(fulfilledObligation(adjustmentPeriodKey))).map(_.obligationDetails)
+      val obligationDetails = obligations(Seq(fulfilledObligation(adjustmentPeriodKey))).flatMap(_.obligationDetails)
```
ian-hmrc: Maybe I missed this from my refactoring, but this can just be `Seq(fulfilledObligation(adjustmentPeriodKey))`. There's no need to wrap it all up as a full response and then unpack it to just the `ObligationDetails`.

test/controllers/returns/submit/adjustments/SelectAdjustmentPeriodControllerSpec.scala
```diff
@@ -40,9 +40,9 @@ class SelectAdjustmentPeriodControllerSpec extends SpecBase with MockitoSugar {
       val obligationDetails = obligations(Seq(
         fulfilledObligation(october2027),
         fulfilledObligation(december2027)
-      )).map(_.obligationDetails)
+      )).flatMap(_.obligationDetails)
```
ian-hmrc: Again, these obligations can be generated more directly rather than wrapping with `obligations()` and then unwrapping with `flatMap()`.

test/controllers/returns/submit/spoilt/SelectSpoiltPeriodControllerSpec.scala
```diff
@@ -82,9 +82,9 @@ class SelectSpoiltPeriodControllerSpec extends SpecBase {
     "must return OK and the correct view when no year parameter is provided" in {
       val mockService = mock[ObligationService]
       val obligationsResponse = createMultiYearObligationsResponse()
-      val obligationDetails = obligationsResponse.obligation.map(_.obligationDetails)
+      val obligationDetails = obligationsResponse.obligation.flatMap(_.obligationDetails)
```
ian-hmrc: Here the full response is being fetched and then filtered down to just the `obligationDetails`. We could just build a list of `ObligationDetails` to start with and get to the result more directly.

I've played with this a bit and it has reduced a lot. I can show you Monday if you like.

test/controllers/returns/submit/adjustments/AdjustmentVolumeWithTypeControllerSpec.scala
ian-hmrc: In the final test in this file, we've lost the obligation for `november2027`, even though an adjustment is being made for that period. The test does pass like this, but the scenario is not realistic. Would it be better to simply have:
```
val obligationDetails = Seq(fulfilledObligation(adjustmentPeriodKey), fulfilledObligation(november2027))
```

test/controllers/returns/submit/spoilt/SelectSpoiltPeriodControllerSpec.scala
ghousedev: Have simplified this, let me know if you had anything else in mind.

ian-hmrc: That looks pretty similar to my noodling. I think the only things extra I have are:
* Extracting `PeriodKey("25AI")` to `september2025`
* Renaming `obligationDetails` to `obligationsSpanning3Years`

ian-hmrc (approved):
LGTM :)

PR #148, opened 2026-08-25
Vpd 1973 - View Payments Page content update

craigemmerson:


app/views/payments/ViewPaymentsView.scala.html
```diff
@@ -28,46 +29,65 @@
...
 @(vm: ViewPaymentsViewModel)(implicit request: Request[_], messages: Messages)
 
-@layout(pageTitle = titleNoForm(messages("payments.viewPayments.title")), fullWidth = true) {
+@unallocatedDetailsContent = {
+    @p(Seq(messages("payments.viewPayments.unallocated.details.intro")))
+    @p(Seq(messages("payments.viewPayments.unallocated.details.includes")))
+    <ul class="govuk-list govuk-list--bullet">
+        <li>@messages("payments.viewPayments.unallocated.details.bullet1")</li>
+        <li>@messages("payments.viewPayments.unallocated.details.bullet2")</li>
+    </ul>
```
ghousedev: Is it worth using the list component we have here?

app/viewmodels/payments/ViewPaymentsViewModel.scala
```diff
@@ -44,7 +44,7 @@ object ViewPaymentsViewModel {
     val clearedYear = returnsDateUtils.getYear.toString
 
     ViewPaymentsViewModel(
-      totalOwed = CurrencyFormatter.currencyFormat(totalOwed),
+      totalOwed = CurrencyFormatter.currencyFormatTwoDp(totalOwed),
```
ian-hmrc: This is the heading and not in a table. This should retain the old formatting.

app/utils/CurrencyFormatter.scala
```diff
@@ -21,7 +21,9 @@ import scala.math.BigDecimal.RoundingMode
 
 trait CurrencyFormatter {
   def currencyFormat(amt: BigDecimal): String = f"£$amt%,1.2f".replace(".00","")
```
ian-hmrc: Could we reduce the duplication here by changing this to `currencyFormatTwoDp(amt).replace(".00","")`

app/utils/CurrencyFormatter.scala
```diff
@@ -21,7 +21,9 @@ import scala.math.BigDecimal.RoundingMode
...
-  
+
+  def currencyFormatTwoDp(amt: BigDecimal): String = f"£$amt%,1.2f"
```
ian-hmrc: Maybe this could be renamed as `currencyFormatInTable()` or similar? Might help distinguish it's use. I think the explanatory comment could be useful too.

app/viewmodels/payments/ViewPaymentsViewModel.scala
ian-hmrc: Cool. I can see that has been reverted.

ian-hmrc (approved):
LGTM :)

